PR #60, opened 2018-06-17
Add Scoop installation method to the README

Calinou:
`usql` was recently added to [Scoop](https://scoop.sh)'s main bucket, which makes it easier/faster to install `usql` on Windows.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,10 +30,11 @@ for other databases.
 
 ## Installing
 
-`usql` can be installed by [via Release][], [via Homebrew][], or [via Go][]:
+`usql` can be installed by [via Release][], [via Homebrew][], [via Scoop][] or [via Go][]:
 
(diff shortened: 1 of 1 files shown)


kenshaw:
I appreciate the contribution. I don't really use Windows, how does `scoop` compare to `chocolately`? Originally I had intentions of writing the .choco script, but it was a giant pain, and more trouble than it was worth.

Calinou:
> I don't really use Windows, how does scoop compare to chocolately? 

See [Chocolatey Comparison](https://github.com/lukesampson/scoop/wiki/Chocolatey-Comparison) on the Scoop wiki :smiley:

majkinetor:
Here is a chocolatey package:

* https://chocolatey.org/packages/usql

It is automatic package that  will follow your updates here within single day

PR #64, opened 2018-06-29
WIP: Add driver 'mylogin'

dolmen:
See https://github.com/dolmen-go/mylogin-driver

Dependency: xo/dburl#7 (submitted, **waiting for merge**, this is the reason this PR is marked "WIP")

Clean build verified with `go build -tags "mylogin no_mysql no_cockroachdb no_memsql no_mssql no_postgres no_redshift no_sqlite3 no_tidb no_vitess"`.


kenshaw:
I appreciate what you are trying to do here (adding support for mylogin.cnf), however I will not be adding this PR. I would suggest you ask the MySQL driver project to support the mylogin.cnf and/or submit a PR to that project in lieu of this.

My rationale for not adding this:

1) It is not tenable to add every *variant* of every Go SQL driver. The multiple MySQL and PostgreSQL drivers that `usql` supports is because both of those databases have more than one, independently built  driver implementations in Go.

2) `usql` already supports a password feature in the `.usqlpass` file, thus adding this would not serve to make `usql` better (ie, it already has this functionality).

3) Even if I was going to accept something like this (I'm not), I would not accept this implementation because none of what you wrote here is actually necessary. You could have changed the mysql driver in `usql` to simply unregister the scheme in `dburl` and then registered a different driver scheme instead. It would have been 2 or 3 lines at maximum. **Please note**, even if you change your PR it will not be accepted to the `usql` project.

Again, if you're intent on getting the mysql driver to support mylogin.cnf out of the box, then I would suggest you submit a PR to that project.

PR #84, opened 2019-01-22
Allow output as JSON

### LINKED ISSUE #37: [Feature Request] Output as json
Is it possible to output rowsets as json instead of a table?

It would be useful when running cli statements piping to other commands.

mtxr:
Fixes #37.


mtxr:
Hi @kenshaw .

1. Redirect to STDERR is ok to change it back. But it doesn't feel right to see anything else than query results when piping to some other command.
2. I thought you would like to create different formatters, that's why I've forked. What do you suggest?

Also, in a short term, could you help me to build binaries like you do for multi platforms? I trying to poroof a concept and I would like to use usql to solve https://github.com/mtxr/vscode-sqltools/issues/13 and https://github.com/mtxr/vscode-sqltools/issues/51. Since my project targets to use many SGDBs as possible, I can keep contributing to usql, I just need to have it distributed for linux/win/darwin, and for x32 and x64.

Thanks.

mtxr:
@kenshaw any updates on this?

How can I contribute?

kenshaw:
@mtxr I appreciate the work, but this deviates too far from where I want `usql` to go, which is a better (but compatible) version of `psql`. This kind of functionality needs to be handled by a `\pset` compatible meta command that can also be enabled via the command line flag (`--pset=VAR[=ARG]`). I don't want to introduce a command line option that is incompatible with `psql` that people then expect to exist in the future, for (eventually) when `usql` does support it via `--pset` and `\pset`.

While `psql` doesn't have "json" formatting, I am 100% perfectly OK with `\pset format json` as an option. Since there is a `psql` direct command line option for enabling HTML (`-H`), I would be fine with a `-J` command line option that also enables JSON. However, this also needs to be configurable via the same meta command semantics.

For what it's worth, someone else (not me) is also working on changing / setting up better output support.

kenshaw:
I've updated `usql` and `tblfmt` in the way I had always meant for the various formatting meta commands to be implemented. I expect there will need to be at least 4 `tblfmt` Encoders, namely: Table, JSON, CSV, Template. Template would be for handling arbitrary Go templates that fulfill the other psql standard output types: html, ms-troff, etc.

I had always expected to pass a map[string]string into the `tblfmt` package, and have it interpret the settings itself.

kenshaw:
Again, I really do appreciate the work you did; however I quickly put together a primitive version of JSON encoding in the way that I had originally intended to implement. If you can please checkout the latest `usql` and `tblfmt` packages, and give it a try, I'd greatly appreciate. If you'd like to contribute further to the output formatting for `usql`, please submit PRs to the `tblfmt` project. Thanks!

PR #113, opened 2020-01-16
Handle pset commandline arguments (fixes #97)

### LINKED ISSUE #97: No difference between -J and -C option
Both -J and -C produces CSV format data instead of JSON and CSV separately in 0.7.5

jetzerb:
The code that handled pset-based commandline arguments using
`kingpin`'s `PreAction` callback was not working as intended.
The context variable passed to the callback contains
information on _all_ commandline arguments, and [doesn't
provide any information specific to the particular flag being
processed](https://github.com/alecthomas/kingpin/issues/125#issuecomment-237856690).

There was also a problem with the callback construction because of
[the way go implements closures](https://golang.org/doc/faq#closures_and_goroutines).
The result is that the variables reflect the last value from the `for`
loop, which is why, as pointed out in #97, there is no difference
between the `-J` and `-C` options: the last iteration of the `for` loop
was for the CSV entry, so regardless which of the `-A, -F, -H, -R, -t,
-T, -x , -z, -0, -J, or -C` options are specified, they all end up
setting `format=csv`.

A new custom type `pset` was created to handle the string-based flags,
and the existing for loop has been modified to properly handle the
boolean flags.

### DIFF
--- a/args.go
+++ b/args.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"strings"
 
 	"github.com/alecthomas/kingpin"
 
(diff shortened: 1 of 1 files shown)


kenshaw:
Much appreciated!

PR #127, opened 2020-09-17
Small readme update for brew install

rothgar:
Adding clarification that `brew` works on Linux and you don't need to `tap` the repo first. Brew can tap and install in the same step.


kenshaw:
Merged.

PR #134, opened 2021-01-17
Exit with non-zero code on query errors

### LINKED ISSUE #69: error exit code
if usql meet error, should return exit -1 or stderr

for example:
x.sql -- wrong sql
```
insert abc; 
123abc
```

`usql mysql://root:pwd@127.0.0.1/test_db -f x.sql`

 only returns

```
Connected with driver mysql (5.7.23)
```

nineinchnick:
Exit with non-zero code on query errors. Makes `handler.Run()` return last command or query error or `nil` if there were none. It now never returns `io.EOF`. `handler.Run()` is the only function that prints errors, so skip printing in `main.go` if a handler error is detected.

Tested with:
```bash
# current master
% usql -W 'pg://postgres@127.0.0.1:5432/test?sslmode=disable' -c "insert into test values (1, 'b')"
Connected with driver postgres (PostgreSQL 13.0 (Debian 13.0-1.pgdg100+1))
error: pq: 23505: duplicate key value violates unique constraint "test_pkey"
% echo $?
0

# this branch
% ./build/linux/0.7.9/usql -W 'pg://postgres@127.0.0.1:5432/test?sslmode=disable' -c "insert into test values (1, 'b')"
Connected with driver postgres (PostgreSQL 13.0 (Debian 13.0-1.pgdg100+1))
error: pq: 23505: duplicate key value violates unique constraint "test_pkey"
% echo $?
1
```

Closes #69

### DIFF
--- a/drivers/errors.go
+++ b/drivers/errors.go
@@ -47,6 +47,9 @@ func (e *Error) Error() string {
 	return e.Driver + ": " + chop(e.Err.Error(), e.Driver)
 }
 
+// Unwrap returns the original error
+func (e *Error) Unwrap() error { return e.Err }
+
 // chop chops off a "prefix: " prefix from a string.
 func chop(s, prefix string) string {
 	return strings.TrimLeftFunc(strings.TrimPrefix(strings.TrimSpace(s), prefix+":"), unicode.IsSpace)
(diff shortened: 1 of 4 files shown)


kenshaw:
Merged.

PR #135, opened 2021-01-17
Add timing option for queries

### LINKED ISSUE #126: Are there plans to implement `\timing` command?
One of the useful commands in `psql` is `\timing` to show query times. Are there any plans to implement it?

nineinchnick:
Add timing option for queries and commands.

In `psql` it looks like:
```
127.0.0.1 postgres@postgres=# \timing on
Timing is on.
127.0.0.1 postgres@postgres=# select count(*) from generate_series(1,100000000);
┌───────────┐
│   count   │
├───────────┤
│ 100000000 │
└───────────┘
(1 row)

Time: 32113.135 ms (00:32.113)
```

In `usql` it would be:
```
pg:postgres@127.0.0.1/test=> \timing
Timing is on.
pg:postgres@127.0.0.1/test=> select count(*) from generate_series(1,100000000);
 count     
-----------
 100000000 
(1 row)

Time: 32851.439 ms (32.851s)
```

I intentionally used the default Go's duration format for >1s to make the code simpler. I believe it serves the purpose of displaying the time in a human-readable way and a deviation from psql like that is acceptable.

Closes #126

### DIFF
--- a/README.md
+++ b/README.md
@@ -599,6 +599,7 @@ Connection
 Operating System
   \cd [DIR]                       change the current working directory
   \setenv NAME [VALUE]            set or unset environment variable
+  \timing [on|off]                toggle timing of commands
   \! [COMMAND]                    execute command in shell or start interactive shell
 
 Variables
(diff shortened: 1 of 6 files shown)


kenshaw:
Merged.

PR #164, opened 2021-01-20
Make prompt work in includes

nineinchnick:
Make prompt work in includes by using same IO in the include handler as in the main handler. Only the statement struct is using the included file IO.

Also, disable password prompts in non-interactive mode.

### DIFF
--- a/contrib/postgres/init.sql
+++ b/contrib/postgres/init.sql
@@ -1,12 +1,12 @@
 \set POSTGRES_USER postgres
 \set POSTGRES_PASS P4ssw0rd
 \set POSTGRES_DB   postgres
-\set POSTGRES_HOST `docker port postgres 5432`
+\set POSTGRES_HOST `docker port postgres 5432 | head -n1`
 
 \prompt NAME 'Create database user: '
 \prompt -password PASS 'Password for "':NAME'": '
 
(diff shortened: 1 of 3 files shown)


kenshaw:
This breaks output when someone has a `.usqlrc` file:

```sh
$ ./usql 
Type "help" for help.

Type "help" for help.

(not connected)=> ^C
(not connected)=> \q
$ rm ~/.usqlrc
$ ./usql 
Type "help" for help.

(not connected)=>  
```

I'll look into fixing this to see if there's an easy workaround. Still investigating if this fixes the other issues with the `\prompt` command.

kenshaw:
@nineinchnick I've merged this, and made some quick minor edits to make the output consistent. Additionally, I'll endeavor to try to fix the output issues and make sure `\prompt` works properly with `\i` and `\ir`. Will not cut a new release until this issue is fixed, as it breaks interactive output.

kenshaw:
@nineinchnick I'm backing out part of this change. With the upcoming `rline` implementation, this will end up getting fixed as a side effect. While `\prompt` now won't work in non-interactive settings, I'd rather have that be the behavior temporarily. After I finish the `rline` implementation, the next push will aim for `psql` compatibility. Planning on getting this done before next major release.

Issue #149, opened 2021-01-18
Implement introspection metacmds (\d*, others)


kenshaw:
All the various introspection metacmds need to be implemented.

For PostgreSQL, partial work [has been done here](https://github.com/xo/pgdesc), however it is not likely that the code in its current state will be completed, as a `regexp` based conversion is not tenable in the long run. While that code can be salvaged, an ideal implementation would be to extract the pertinent functions from the postgres source, then transform that using clang's `--ast-dump` (to json). Then, use Go to traverse the generated C AST and generate the necessary code.

Additionally, the other 4 major databases (SQLite3, MySQL, MS SQL, Oracle) need to have introspection commands implemented, which likely means painstaking and time consuming research to determine the correct. Since MySQL is open source and supports similar commands as `psql` in its command-line client, it may be possible to also write a generation script for it as well. A separate, generic `information_schema` one should also be written to be used as default for the non major databases.

nineinchnick:
Can we close this now and keep #57 for 100% compatibility with psql?

Issue #110, opened 2020-01-07
Context based auto completion?


pricheal:
The overview section of the readme specifices that this tool supports context based auto completion, but I can't find anything on that. Does usql currently support any type of completion, and if it doesn't, are there any plans to do so?

nineinchnick:
We plan on working on that right after database introspection commands (`\d*`) in #57  and #149. Note that #150 was created recently which is a dup of this issue.

Issue #57, opened 2018-06-15
Postgres \dt


stuaxo:
IIRC \dt and friends are implemented in the psql client ...  still, psql users will be familiar with it ... so I guess this is about adding per-database commands.

As it is I'm not sure how to list the tables, so probably go back to psql for now.

kenshaw:
Yes, sometime before 1.0, this will be supported by `usql`. I do not have a specific ETA on that, as this is a lot of work. For all the primary databases that `usql` supports, (mysql, postgres, oracle, mssql, sqlite3), this would need to be implemented.

dansmill:
Im very excited about this project and can see it being my main interface for sure with all the different databases that I use, but until \d family is implemented I will just keep my eye on this project.  I use them way to much.  Keep up the great work!

kenshaw:
I have made some progress towards this. It's not done (nor will be done) until this support works for the 5 major, primary databases that `usql` targets (PostgreSQL, MySQL, Oracle, Microsoft SQL, SQLite3). Additionally, I have made the constraint that I will not release this until the logic for `\d*` is the same for `usql` as it is for `psql`. I spent a lot of time about 7 months ago writing a simple, targeted translator for PostgreSQL's `describe.c` to Go, which is available: https://github.com/xo/pgdesc but is only 80% complete at this time. There are only so many hours in the day, and limited time that I can dedicate to this. 

Hypothetically, if `usql` were the only code base I was working on, most of this could have been done well over a year ago.

nineinchnick:
@kenshaw I'd like to give this one a shot. Do you have something in progress or can I start diving in?

kenshaw:
@nineinchnick If this feature gets implemented, I really would want "100%" compatibility with `psql`. The way `psql` works is that the `\d` (and other informational) queries are part of the `psql` code, not the server code. As such, I started `github.com/xo/pgdesc` and had started writing some regexp based translation scripts for postgres' direct source, which I realized would not work as the logic is just too complicated to translate with regexp replacement. I really would want this generated from the postgres source at minimum, since it changes with every major postgres revision and would want to continue generating it.

[...]

Issue #41, opened 2018-03-15
[Feature Request] Autocomplete


gabrielps:
It would be nice to see some autocomplete when introducing files after \i
I would ask for autocomplete sql keywords, table names and column names (as in psql), but it seems like a lot to do :smile:

Leask:
vote

PR #167, opened 2021-02-02
Basic \d* implementation

nineinchnick:
This is an initial implementation of the describe (\d) commands. The goal is to settle on interfaces and provide examples, not to be feature complete.

There's a new metadata package that has two main interfaces:
* reader - for reading metadata into structs like Schema, Table, Column, Index, IndexColumn, etc.
* writer - for writing a human-readable description of selected metadata

Reader is based on [JDBC](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html). There's the `informationschema` package, implementing a reader based on the tables from `information_schema` schema. Since this schema has differences across various DBs, it requires options to configure its behavior.

Writer is based on [psql's describe.c](https://github.com/postgres/postgres/blob/master/src/bin/psql/describe.c). Writer implementations can use readers, but they don't have to. There's a `DefaultWriter` that does use the reader.

Both reader and writer are configured for 3 databases:
* PostgreSQL, with a custom reader for databases and indexes
* MySQL
* Trino, with a custom writer for catalogs (`listAllDbs` aka `\l`)

There's full test coverage! Well, over 70% at least. Those are integration tests, running on real databases in containers. Tests are simple, they just compare output saved in text files (expected vs actual). Easy to extend, easy to manage changes. There's also output from native clients, to see how much we're behind (psql) or how much usql is already better (mysql).

Things to consider during the review, in order:
* Column's type does NOT include scale and precision, these are in separate properties; I wanted to avoid adding logic based on data type names in this first iteration, but it makes output harder to read
* function types don't work well, `information_schema` doesn't distinct aggregates
* there's a bunch of TODO comments left in the code
* should PostgreSQL custom reader be moved into a separate package, under `driver/metadata`? It would be easier to reuse outside of `usql`
* parsing of `[S+]` options for every `\d` command is very basic, intentionally
* pattern matching is NOT using regexes, like in psql, only SQL wildcards, this is intentional
* table formatting will be improved in separate PRs (centered titles, extra footers, pager support etc.)
* not all exported functions have proper comments - how bad we want this?
* consider if readers will be useful for autocomplete #110  and #41 
* in `ListIndex` and `DescribeFunctions`, `IndexColumns` and `FunctionColumns` are fetched in separate queries for every index and function, instead of using string aggregates, since these vary across DBs - do we want more options in `informationschema` reader for that?

Out of scope (next PRs) - should be turned into issues when this is merged:
* relations, FKs, constraints, triggers in `\d` (`describeTableDetails`)
* extra details for sequences, indexes, views (in footers)
* describe roles
* implement for MSSQL, Oracle, SQLite
* test more database versions, currently only the latest is tested

Bug: #57 and #149


drivers/metadata/metadata_test.go
```diff
@@ -0,0 +1,299 @@
...
+	if err != nil {
+		log.Fatalf("Could not connect to docker: %s", err)
+	}
+
+	for _, db := range dbs {
+		var ok bool
+		db.Resource, ok = pool.ContainerByName(db.RunOptions.Name)
+		if !ok {
+			buildOpts := &dt.BuildOptions{
+				ContextDir: ".",
```
nineinchnick, line 119: Missing `Dockerfile` in this package.

kenshaw:
Merged.

Issue #169, opened 2021-02-06
Add system schemas option to informationschema reader


nineinchnick:
Turns out that there's a `SYSTEM TABLE` type, so a list of system schemas (like `information_schema` or `pg_catalog`) should be passed to the reader. Filtering can still be done in the writer to avoid changing the reader's API.

Ref: https://github.com/mysql/mysql-connector-j/blob/d64b664fa93e81296a377de031b8123a67e6def2/src/main/user-impl/java/com/mysql/cj/jdbc/DatabaseMetaDataUsingInfoSchema.java#L798

nineinchnick:
Uh I don't know why I created this. If the reader was supposed to return tables of the `SYSTEM TABLE` type, the underlying DB would do so.

PR #178, opened 2021-02-06
SQLite db metadata reader

### LINKED ISSUE #175: Implement \d commands for SQLite.
Implement \d commands for SQLite.

Would be nice to implement only a metadata reader and use the default writer.

Ref: https://github.com/xerial/sqlite-jdbc/blob/6ca62fd07bf7faf58e923ba8651757621ffc693c/src/main/java/org/sqlite/jdbc3/JDBC3DatabaseMetaData.java

nineinchnick:
Implement SQLite db metadata reader using `sqlite_master` and `PRAGMA` statements. This includes:
* tables, with columns
* functions
* indexes, with columns
* schemas (virtual tables) - since SQLite's SQL syntax does refer to them as `<schema>.<table>`

`sqlite_master` also contains a list of triggers, but the reader API doesn't yet have triggers defined.

This PR also fixes locations of the Dockerfile used in all metadata related tests.

Minor unrelated bugfixes include:
* define more table types, like `SYSTEM TABLE` - this is related to #169 
* avoid filtering functions by type, if not requested
* don't print schema name when empty in table titles

For tests, I used the official Docker Go SDK. Turns out the `ory/dockertest` module uses a copy of some other client. We can migrate away from it in future PRs.

These tests only use a container to fetch the example schema. This might be overkill, but I wanted to avoid committing it to the repo, creating yet another copy.

Closes #175


kenshaw:
Merged.

PR #180, opened 2021-02-07
Oracle db metadata reader

### LINKED ISSUE #153: Implement SHOW_HIDDEN
Once the `\d*` and other introspection inquiries are available, need to implement `SHOW_HIDDEN` to have similar output to `psql` for compatibility purposes.

### LINKED ISSUE #174: Implement \d commands for Oracle
Implement \d commands for Oracle.

Would be nice to implement only a metadata reader and use the default writer.

This ODBC driver might have a similar API to the current reader interfaces: https://github.com/nextgres/oss-edo-odbc-driver-for-oracle/blob/master/Sources/MetaData.cpp

nineinchnick:
Implement Oracle metadata reader. Since it's used by two drivers, I've put it into a standalone module.

Since there's a ton of schemas, and Oracle's catalogs are terribly slow to read, I tried implementing dynamic filtering to skip all of them except the one matching the current user. This has a somehow surprising effect where you don't get any results unless you use commands like `\dtS sometable%`. I thought it's an issue with binding params, lost 3 hours staring at the screen, and implemented support for `ECHO_HIDDEN` setting. I went ahead and added it for all readers, since it will be _very_ useful for debugging.

We need to circle back around filtering out system objects. Doing this in the writer is ok-ish for most databases, but too late or Oracle. I should probably remove it for Oracle completely.

Closes #174  and closes #153

### DIFF
--- a/drivers/drivers.go
+++ b/drivers/drivers.go
@@ -85,9 +85,9 @@ type Driver struct {
 	// BatchQueryPrefixes will be used by BatchQueryPrefixes if defined.
 	BatchQueryPrefixes map[string]string
 	// NewMetadataReader returns a db metadata introspector.
-	NewMetadataReader func(db DB) metadata.Reader
+	NewMetadataReader func(db DB, opts ...metadata.ReaderOption) metadata.Reader
 	// NewMetadataWriter returns a db metadata printer.
-	NewMetadataWriter func(db DB, w io.Writer) metadata.Writer
+	NewMetadataWriter func(db DB, w io.Writer, opts ...metadata.ReaderOption) metadata.Writer
 }
 
 // drivers is the map of drivers funcs.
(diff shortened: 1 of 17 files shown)


nineinchnick:
I just found out that Trino considers these schemas to be internal:
```
            .add("ctxsys")
            .add("flows_files")
            .add("mdsys")
            .add("outln")
            .add("sys")
            .add("system")
            .add("xdb")
            .add("xs$null")
```

nineinchnick:
Ok I made the list of Oracle's system schemas constant.

kenshaw:
I wasn't able to merge this change, as it conflicts drastically with the other metadata ones. You seem to have 2 different APIs in play here?

nineinchnick:
> I wasn't able to merge this change, as it conflicts drastically with the other metadata ones. You seem to have 2 different APIs in play here?

I was expecting conflicts because I also implemented `ECHO_HIDDEN` in this PR. At least it's a separate commit at least. Anyway, I resolved them and It should be ready for merging now.

PR #187, opened 2021-03-20
Metadata writer should print aligned table titles

nineinchnick:


nineinchnick:
I'll merge this after fixing https://github.com/xo/tblfmt/issues/15 to also bump tbmlfmt version.

PR #189, opened 2021-03-20
Basic context-based autocomplete

### LINKED ISSUE #110: Context based auto completion?

### LINKED ISSUE #150: Implement context-based completion
After introspection and a switching to a readline package is completed, context-based completion would be simple/straightforward, not only for SQL statements, but also for table, fields, et al.

nineinchnick:
Add a pretty basic, generic context-based autocomplete, inspired heavily by psql's [tab-complete.c](https://github.com/postgres/postgres/blob/master/src/bin/psql/tab-complete.c). Currently, only a small subset of completions from psql are implemented, since most will be PostgreSQL specific. This first iteration is focused on SELECT, INSERT, UPDATE, DELETE.

There's a new interface allowing drivers to provide their own completer and the default one should be a fallback. If it won't work well for some drivers, due to the reader not being able to read the catalog quickly enough (👀 Oracle), an empty struct can be passed as a reader, falling back to completing only SQL keywords. I don't think we need a way to disable it completely.

The completer will execute multiple database queries against system catalogs. There's no caching, on purpose.

I've NOT tested it with Oracle, only PostgreSQL, SQLServer, and SQLite. There are basic unit tests.

~This PR includes the commit from #188, so either individual commits should be reviewed, or #188 needs to be merged first.~

Closes #150  and closes #110. Doesn't yet solve #41, which asks for autocomplete of files after `\i`, but it would be trivial to add after this.

If this gets merged, I'll create issues for the next steps that include:
* make the completer extendable to easily configure SQL keywords supported by particular databases, start with SQLite, MySQL, MSSQL
* complete arguments for backslash commands, including listing local files for #41
* ~add more contexts for DML commands like INSERT, UPDATE, DELETE~
* add more contexts for DDL commands like CREATE, ALTER

### DIFF
--- a/drivers/completer/completer.go
+++ b/drivers/completer/completer.go
@@ -1,0 +1,749 @@
+// completer package provides a generic SQL command line completer
+package completer
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"sort"
+	"strings"
+	"unicode"
+
+	"github.com/gohxs/readline"
+	"github.com/xo/usql/drivers/metadata"
+	"github.com/xo/usql/env"
+)
+
+const (
+	WORD_BREAKS = "\t\n@$><=;|&{() "
+)
+
+type caseType bool
+
+var (
+	IGNORE_CASE = caseType(true)
+	MATCH_CASE  = caseType(false)
+)
+
+func NewDefaultCompleter(r metadata.Reader) func(db metadata.DB) readline.AutoCompleter {
+	return func(db metadata.DB) readline.AutoCompleter {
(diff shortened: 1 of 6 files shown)


nineinchnick:
I guess I can add INSERT, UPDATE and DELETE completions to this PR. It won't make it much bigger but much more complete (ha!).

PR #200, opened 2021-03-29
Enable metadata reader for Snowflake

nineinchnick:
Tested manually, requires to have a database and warehouse selected, using `use database` and `use warehouse`.

Tables names are stores in uppercase. We might want to handle this, but I don't think we need to in this PR.

There's actually a `information_schema.databases` table/view available so maybe it's a good moment to implement (an optional, disabled by default) `Catalogs()`?

### DIFF
--- a/drivers/snowflake/snowflake.go
+++ b/drivers/snowflake/snowflake.go
@@ -24,6 +24,9 @@ func init() {
 	gosnowflake.SetLogger(&l)
 	newReader := infos.New(
 		infos.WithPlaceholder(func(int) string { return "?" }),
+		infos.WithCustomColumns(map[infos.ColumnName]string{
+			infos.SequenceColumnsIncrement: "''",
+		}),
 		infos.WithFunctions(false),
 		infos.WithIndexes(false),
 	)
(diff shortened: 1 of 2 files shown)


nineinchnick:
We can also consider creating a generic metadata writer (not reader!) that would try executing `SHOW` queries. This would make the output of `\d` commands more consistent for users that _are_ familiar with using `SHOW`.

drivers/snowflake/snowflake.go
```diff
@@ -4,26 +4,48 @@
...
 
 func init() {
 	r := logrus.New()
 	r.Out, r.Level = ioutil.Discard, logrus.PanicLevel
 	var l gosnowflake.SFLogger = &logger{r}
 	gosnowflake.SetLogger(&l)
+	newReader := infos.New(
+		infos.WithPlaceholder(func(int) string { return "?" }),
+		infos.WithCustomColumns(map[infos.ColumnName]string{
+			infos.SequenceColumnsIncrement: "NULL",
```
nineinchnick, line 28: I don't think we can use null here, it'll fail to be scanned into a regular non-pointer string.

PR #219, opened 2021-04-07
Handle SIGINT during opening a connection

### LINKED ISSUE #201: Handle signals
Handle signals like `ctrl+c` to interrupt the current query, instead of terminating `usql`.

nineinchnick:
I noticed that when running the `\c` command with an invalid host, when I press `CTRL+C` then `usql` quits. This PR adds a context that's being closed on SIGINT.

Unfortunately, the `Open()` function from `database/sql` package doesn't use a context. We do call `Ping()` right after it, so I'm passing the context there. But some drivers, notably `libpq`, perform blocking operations when calling `Open()`, like dialing. This can block for the duration of `connect_timeout`. Adding a context to handle SIGINT means `usql` cannot be interrupted at all. `pgx` behaves correctly. Can I suggest that we make `pgx` the default driver for `postgres://`, especially since `libpq` is in maintenance mode?

Another solution is to also pass the context to the `driver.Open` callback, define `Open()` for the `postgres` driver and make it panic when context gets closed.

Closes #201

### DIFF
--- a/drivers/athena/athena.go
+++ b/drivers/athena/athena.go
@@ -4,6 +4,7 @@
 package athena
 
 import (
+	"context"
 	"regexp"
 
 	_ "github.com/uber/athenadriver/go" // DRIVER: athena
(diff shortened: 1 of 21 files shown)


kenshaw:
Merged.

PR #229, opened 2021-04-16
Implement quiet setting

### LINKED ISSUE #221: Recognize `-q` flag and the `QUIET` variable
From psql manual:
> -q
> --quiet
>
>    Specifies that psql should do its work quietly. By default, it prints welcome messages and various informational output. If this option is used, none of this happens. This is useful with the -c option. This is equivalent to setting the variable QUIET to on.

nineinchnick:
There are slight differences with how it's implemented in `psql`. For example, `QUIET` can be unset in `usql`, but it's set to `off` in `psql`.

Closes #221

### DIFF
--- a/args.go
+++ b/args.go
@@ -133,6 +133,10 @@ func NewArgs() *Args {
 			return nil
 		}).Bool()
 	}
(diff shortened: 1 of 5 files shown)


nineinchnick:
Merged in 7714c9dc25e346d59c5c5f02c6adddb7d4ae35d8

Issue #166, opened 2021-01-21
Fix quoted string processing and variable interpolation


kenshaw:
Quoted strings, variable interpolation, are not quite compatible with `psql` in metacmds.

I've added `testdata/quotes.sql` that demonstrate the various ways  that `psql` works that `usql` does not. 

The output for `psql`:
```sh
-- conditional variables display FALSE when name is not set
\unset foo
\echo :{?foo}
FALSE
-- conditional variables display TRUE when name is set
\set foo 'bar'
\echo :{?foo}
TRUE
-- single quoted strings will decode '' as ' and decode \n, \t, \b, \r, \f, \digits octals, \xdigits (standard escapes)
\set foo 'bar''bar\r\n'
-- single quoted variables escape ' but does not escape special characters
\echo :'foo'
'bar''bar
'
-- double quoted variables do not escape ' or special characters
\echo :"foo"
"bar'bar
"
-- single quoted strings decode any other standard escape (\<char>) as literal
\set foo 'bar\'''bar'
\echo :foo
bar''bar
\echo :'foo'
'bar''''bar'
-- single quoted variables escape \ with E'' style strings
\set foo 'bar\\\''
\echo :foo
bar\'
\echo :'foo'
 E'bar\\'''
\echo :"foo"
"bar\'"
-- backticks interpolate unquoted variables
\set foo 'bar'
\echo `echo :foo`
bar
-- backticks interpolate single quoted variables
\echo `echo :'foo'`
bar
-- backticks do not interpolate double quoted variables
\echo `echo :"foo"`
:foo
-- backticks have error messages for single quoted variables containing \r or \n when using :'' syntax
\set foo 'bar\r\n'
\echo `echo :'foo'`
shell command argument contains a newline or carriage return: "bar
"
:foo
```

And current `usql`:
```sh
-- conditional variables display FALSE when name is not set
\unset foo
\echo :{?foo}
:{?foo}
-- conditional variables display TRUE when name is set
\set foo 'bar'
\echo :{?foo}
:{?foo}
-- single quoted strings will decode '' as ' and decode \n, \t, \b, \r, \f, \digits octals, \xdigits (standard escapes)
\set foo 'bar''bar\r\n'
-- single quoted variables escape ' but does not escape special characters
\echo :'foo'
'bar'bar\r\n'
-- double quoted variables do not escape ' or special characters
\echo :"foo"
"bar'bar\r\n"
-- single quoted strings decode any other standard escape (\<char>) as literal
[...]

PR #242, opened 2021-04-26
Unescape special chars in params

nineinchnick:
Partially addresses #166 by unescaping special characters, including quotes. This allows to use of variables as aliases for long/complex queries that contain quoted literals, like so:
```
\set activity 'select pid,now()-query_start,substring(query for 60) || \'...\' || substring(query from length(query)-40) from pg_stat_activity where state = \'active\' order by query_start desc;'
```
or use special chars in commands:
```
\echo '\nWelcome, my magistrate\n'
```

This does NOT address any escaping done when quoting vars, per `testdata/quotes.sql`.

### DIFF
--- a/env/env.go
+++ b/env/env.go
@@ -16,7 +16,9 @@ import (
 	"path/filepath"
 	"regexp"
 	"runtime"
+	"strconv"
 	"strings"
+	"unicode/utf8"
 
 	"github.com/xo/dburl"
 	"github.com/xo/usql/text"
@@ -366,18 +368,38 @@ func Exec(s string) (string, error) {
 	return string(buf), nil
 }
 
-var cleanDoubleRE = regexp.MustCompile(`''`)
+var cleanDoubleRE = regexp.MustCompile(`(^|[^\\])''`)
 
 // Dequote unquotes a string.
-func Dequote(s string, c byte) (string, error) {
-	if len(s) < 2 || s[len(s)-1] != c {
+func Dequote(s string, quote byte) (string, error) {
+	if len(s) < 2 || s[len(s)-1] != quote {
 		return "", text.ErrUnterminatedQuotedString
 	}
 	s = s[1 : len(s)-1]
-	if c != '\'' {
-		return s, nil
+	if quote == '\'' {
+		s = cleanDoubleRE.ReplaceAllString(s, "$1\\'")
 	}
-	return cleanDoubleRE.ReplaceAllString(s, "'"), nil
+
+	// this is the last part of strconv.Unquote
+	var runeTmp [utf8.UTFMax]byte
+	buf := make([]byte, 0, 3*len(s)/2) // Try to avoid more allocations.
+	for len(s) > 0 {
+		c, multibyte, ss, err := strconv.UnquoteChar(s, quote)
+		switch {
+		case err != nil && err == strconv.ErrSyntax:
+			return "", text.ErrInvalidQuotedString
+		case err != nil:
+			return "", err
+		}
+		s = ss
+		if c < utf8.RuneSelf || !multibyte {
+			buf = append(buf, byte(c))
+		} else {
+			n := utf8.EncodeRune(runeTmp[:], c)
+			buf = append(buf, runeTmp[:n]...)
+		}
+	}
+	return string(buf), nil
 }
 
 // Getvar retrieves an environment variable.
--- a/stmt/params_test.go
+++ b/stmt/params_test.go
@@ -48,50 +48,53 @@ func TestDecodeParamsGetAll(t *testing.T) {
 	tests := []struct {
 		s   string
 		exp []string
-		err bool
+		err error
 	}{
-		{``, nil, false},
-		{` `, nil, false},
-		{` :foo`, []string{`bar`}, false},
-		{` :'foo`, nil, true},
-		{`:'foo'`, []string{`'bar'`}, false},
-		{`:'foo':foo`, []string{`'bar'bar`}, false},
-		{`:'foo':foo:"foo"`, []string{`'bar'bar"bar"`}, false},
-		{`:'foo':foo:foo`, []string{`'bar'barbar`}, false},
-		{` :'foo':foo:foo`, []string{`'bar'barbar`}, false},
-		{` :'foo':yes:foo`, []string{`'bar':yesbar`}, false},
-		{` :foo `, []string{`bar`}, false},
-		{`:foo:foo`, []string{`barbar`}, false},
-		{` :foo:foo `, []string{`barbar`}, false},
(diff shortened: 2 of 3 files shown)


env/env.go
```diff
@@ -366,18 +368,39 @@ func Exec(s string) (string, error) {
...
+	if quote == '\'' {
+		s = cleanDoubleRE.ReplaceAllString(s, "$1\\'")
+	}
+
+	// this is the last part of strconv.Unquote
+	var runeTmp [utf8.UTFMax]byte
+	buf := make([]byte, 0, 3*len(s)/2) // Try to avoid more allocations.
+	for len(s) > 0 {
+		c, multibyte, ss, err := strconv.UnquoteChar(s, quote)
+		if err != nil {
```
kenshaw, line 389: This should be a switch statement, not nested `if`'s.

env/env.go
```diff
@@ -366,18 +368,39 @@ func Exec(s string) (string, error) {
...
-func Dequote(s string, c byte) (string, error) {
-	if len(s) < 2 || s[len(s)-1] != c {
+func Dequote(s string, quote byte) (string, error) {
+	if len(s) < 2 || s[len(s)-1] != quote {
 		return "", text.ErrUnterminatedQuotedString
 	}
 	s = s[1 : len(s)-1]
-	if c != '\'' {
-		return s, nil
+
```
kenshaw, line 379: Not that long ago, I decided it's better to not include empty lines in Go code:

1. If code requires a visual break, it's better to fill that with an explanatory comment
2. Different individuals will occasionally mess up and put in multiple lines, not format their code, etc., or one person will introduce a line, the next will remove it, and so on. It's not that the people are arguing, but that they just may naturally do that. Since it's impossible to control this from individual to individual due to the impossibility of creating a standard guide on "when" and "where" an empty line to break up logic, I've come to the conclusion that it's just better to not have them anywhere. When/if a person decides it's necessary, they should see 1 
3. Less empty lines means its easier to view more lines at once on screen
4. Reduces commit "chatter" (see 2)

stmt/params_test.go
```diff
@@ -48,50 +48,53 @@ func TestDecodeParamsGetAll(t *testing.T) {
...
 	for i, test := range tests {
 		vals, err := DecodeParams(test.s).GetAll(testUnquote(u, t, i, test.s))
-		if test.err && err != text.ErrUnterminatedQuotedString {
-			t.Fatalf("test %d expected unterminated quoted string error, got: %v", i, err)
+		if err != test.err {
+			t.Fatalf("test %d for %#v expected err %v, got: %v", i, test.s, test.err, err)
 		}
 		if !reflect.DeepEqual(vals, test.exp) {
-			t.Errorf("test %d expected %v, got: %v", i, test.exp, vals)
+			t.Errorf("test %d for %#v expected %v, got: %v", i, test.s, test.exp, vals)
```
kenshaw, line 97: Instead of `%#v` it should be `%q` (same on the one above).

stmt/params_test.go
```diff
@@ -48,50 +48,53 @@ func TestDecodeParamsGetAll(t *testing.T) {
...
-		{` :'foo' :"foo" `, []string{`'bar'`, `"bar"`}, false},
-		{` :'foo' :"foo"`, []string{`'bar'`, `"bar"`}, false},
-		{` :'foo'  :"foo"`, []string{`'bar'`, `"bar"`}, false},
-		{` :'foo'  :"foo" `, []string{`'bar'`, `"bar"`}, false},
-		{` :'foo'  :"foo"  :foo `, []string{`'bar'`, `"bar"`, `bar`}, false},
-		{` :'foo':foo:"foo" `, []string{`'bar'bar"bar"`}, false}, // 30
-		{` :'foo''yes':'foo' `, []string{`'bar'yes'bar'`}, false},
-		{` :'foo' 'yes' :'foo' `, []string{`'bar'`, `yes`, `'bar'`}, false},
-		{` 'yes':'foo':"foo"'blah''no' "\ntest" `, []string{`yes'bar'"bar"blah'no`, `\ntest`}, false},
+		{``, nil, nil},
```
kenshaw, line 53: Not really part of this, but a mental note: these tests are getting unwieldy, and should likely be compared directly to `psql`'s output via functionality similar to how the `\crosstabview` test data is generated in `tblfmt`. That is, separate the strings out into a `*gen.go` file, and then execute against `psql` to generate a static test file. This test then should load that from that separate file, and check/ensure it generates the same output as what `psql` generated.

kenshaw (changes requested):

env/env.go
nineinchnick: I disagree, but I can commit to it. I've noticed that in your commits a while ago but I didn't know the reasons. Please put this info in `CONTRIBUTING.md` so we can reference it in other PRs.

Here are my opinions, but this is just FYI, we don't need to discuss any further:
1. This is subjective, but I don't think a comment is enough of a visual break.
2. Agree here, we can make such rules in `CONTRIBUTING.md`. BTW formatting code using `go fmt` is/should be a hard requirement for any Go project. Also, it doesn't allow more than one empty line :-)
3. Disagree, but this is subjective. I believe you only need the current context to fit on the screen, and this is usually a single function. If it's too big, break it down into smaller functions. Blocks separated with empty lines are natural candidates to be extracted (and named). But as always, "it depends". I guess it's not mentioned in either [Effective Go](https://golang.org/doc/effective_go.html) or [code review comments](https://github.com/golang/go/wiki/CodeReviewComments) since it's so subjective.

stmt/params_test.go
nineinchnick: Thanks!

env/env.go
kenshaw: Eh, it's not something worth spending a millisecond on. I just have a preference now for spaceless code. The reason I got to this is after working on a lot of other projects and seeing developers add/remove so many empty lines, I've come to the conclusion that the empty lines in blocks of code are worthless. The rationale/thinking kind of goes like this: if a block of code is more than 10-15 lines, it "should" be broken up into multiple functions (which I try to follow), and an empty line isn't really enough of a visual break. When code can't be broken up into multiple functions (such as large switch's or something else), and it's more than 10-15 lines, then each "block" should be delineated by a comment. Having _both_ an empty line and a comment is redundant. Again, this isn't worth debating / arguing over -- I'm not going to reject code with superfluous empty lines, I'm just stating that I think it's easier to read/view without the lines, especially when reviewing via patch sets (ie, in Git or some other patch based commit tracker).

stmt/params_test.go
nineinchnick: I agree, I'll try to add this in a follow up PR.

env/env.go
nineinchnick: Done.

PR #272, opened 2021-07-16
Fix "invalid character" error during completion on oracle

### LINKED ISSUE #268: bug: tab completion errors out for oracle (default driver)
Trying to use tab completion within an oracle database gives an error:
```
2021/07/08 00:16:56 Error getting selectables ORA-00911: invalid character
```

Example:
image

jpalus:
'schema' format is used for two cases:
- schema explicitly provided by user and used as a query parameter
- owning user schema used implicitly by calling USER function

Therefore 'schema' format cannot use positional parameter in order to
be able to handle function call. Use string instead and turn it into
positional parameter or call a function.

Fixes #268

### DIFF
--- a/drivers/metadata/oracle/metadata.go
+++ b/drivers/metadata/oracle/metadata.go
@@ -112,7 +112,7 @@ o.object_type AS table_type
 FROM all_objects o
 `
 	conds, vals := r.conditions(f, formats{
-		schema:     "o.owner LIKE :%d",
+		schema:     "o.owner LIKE %s",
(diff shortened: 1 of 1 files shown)


kenshaw:
@nineinchnick I've reviewed this, but since I didn't write any of the metadata parts, I'm not sure if the original intent was to inject a placeholder or to inject the value directly. Your input would be appreciated.

nineinchnick:
This fix is correct, thanks! Sorry it took me so long to get to this. I'll try to add tests there so we can avoid such issues in future.

PR #276, opened 2021-07-18
Make oracle completion case insensitive

jpalus:
oracle stores metadata in uppercase so make sure queried data is also
uppercase

### DIFF
--- a/drivers/metadata/oracle/metadata.go
+++ b/drivers/metadata/oracle/metadata.go
@@ -439,7 +439,7 @@ func (r metaReader) conditions(filter metadata.Filter, formats formats) ([]strin
 	conds := []string{}
 	vals := []interface{}{}
 	if filter.Schema != "" && formats.schema != "" {
-		vals = append(vals, filter.Schema)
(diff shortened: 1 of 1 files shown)


kenshaw:
@nineinchnick Similarly, with this one -- I think we should change all of the metadata queries to be case insensitive. Should probably be done in the actual SQL though and not from the Go side.

jpalus:
Actually Oracle allows case sensitive identifiers but it complicates things. Case sensitive identifiers must be put in quotes, which in turn needs to be recognized by metadata code, trimmed and case sensitive query is done instead. To make it worse completion candidates returned by query need to be wrapped in quotes too.

I'll see whether I can implement it without knowing anything about Go, but perhaps better leave it for another PR.

nineinchnick:
I'm not sure quoting would be an issue. Anyway, this would be a pretty big task, so I created a dedicated issue for this: #282. We need to add some basic tests to the metadata packages first.

I'm ok with merging this one as is.

nineinchnick (approved):

jpalus:
Rebased onto current master.

PR #284, opened 2021-07-30
Added triggers in \d (describeTableDetails)

### LINKED ISSUE #231: Add triggers in \d (describeTableDetails)
* [ ] Add a new metadata reader interface,
* [ ] Implement it in the information_schema reader
* [ ] Use it in the default metadata writer so that `\d` command prints triggers info for tables

jgryko5:
Closes  #231.

### DIFF
--- a/drivers/metadata/postgres/metadata.go
+++ b/drivers/metadata/postgres/metadata.go
@@ -193,6 +193,50 @@ FROM pg_catalog.pg_class c
 	return metadata.NewIndexColumnSet(results), nil
 }
 
+func (r metaReader) Triggers(f metadata.Filter) (*metadata.TriggerSet, error) {
+	qstr := `SELECT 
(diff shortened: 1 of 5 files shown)


drivers/metadata/postgres/metadata.go
```diff
@@ -193,6 +194,58 @@ FROM pg_catalog.pg_class c
...
+	if err != nil {
+		return nil, err
+	}
+	defer closeRows()
+
+	results := []metadata.Trigger{}
+	for rows.Next() {
+		rec := metadata.Trigger{}
+		err = rows.Scan(
+			&rec.Name,
```
nineinchnick, line 235: Since there's filtering by scheme and parent, can you add their names to `SELECT` and fill them here?

nineinchnick (approved):

PR #288, opened 2021-07-31
Add custom completer for MySQL

### LINKED ISSUE #285: Autocomplete MySQL specific keywords like USE <database>
hi, i use mycli before, the code completion seems to be quick, can usql make the improvement of code completion?
great thanks for this amazing tool

nineinchnick:
A custom MySQL completer that completes the `USE <database>` query.

Closes #285

### DIFF
--- a/drivers/metadata/mysql/metadata.go
+++ b/drivers/metadata/mysql/metadata.go
@@ -1,0 +1,66 @@
+package mysql
+
+import (
+	"time"
+
+	"github.com/gohxs/readline"
+	"github.com/xo/usql/drivers"
+	"github.com/xo/usql/drivers/completer"
+	"github.com/xo/usql/drivers/metadata"
+	infos "github.com/xo/usql/drivers/metadata/informationschema"
+)
+
+var (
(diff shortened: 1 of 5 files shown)


drivers/metadata/mysql/metadata.go
```diff
@@ -0,0 +1,58 @@
...
+			metadata.WithLimit(1000),
+		}
+		reader := NewReader(db, readerOpts...)
+		opts = append([]completer.Option{
+			completer.WithReader(reader),
+			completer.WithDB(db),
+			completer.WithSQLStartCommands(append(completer.CommonSqlStartCommands, "USE")),
+			completer.WithBeforeComplete(func(previousWords []string, text []rune) [][]rune {
+				schemaNames := []string{}
+				schemas, err := reader.(metadata.SchemaReader).Schemas(metadata.Filter{WithSystem: true})
```
nineinchnick, line 43: This should only be done either when matched `USE` or before creating the completer, because otherwise a query fetching schemas/databases will be executing _on each tab key press_.

PR #302, opened 2021-11-06
adding missing @master and removing redundant section in README.md. Resolves #293

### LINKED ISSUE #293: USQL Installation instruction - GO GET is deprecated
The instructions for "Installing via Go" , include this statement,
$ GO111MODULE=on go get github.com/xo/usql

And similar like this for adding oracle go driver.
GO111MODULE=on go get -tags 'godror odbc' github.com/xo/usql

When these are run, a warning about 'get' deprecation is given. It's just a deprecation warning, so not critical today, one can ignore it. However at some point, presumably, this won't be supported. 

go get: installing executables with 'go get' in module mode is deprecated.
	Use 'go install pkg@version' instead.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.

Presumably one should instead do this command with "go install" and specify a version #, such as ....

GO111MODULE=on go install -tags 'godror odbc' github.com/xo/usql@latest

But this fails with an ERROR message about replace directives.

go install: github.com/xo/usql@latest (in github.com/xo/usql@v0.9.4):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.

jpy-git:
Fixed a couple of missing `@master`s at the end of the `go install -tags` lines in README.md.

Additionally propose removing 2 of the `go install -tags` lines as they are repeated immediately below.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,16 +124,6 @@ scoop install usql
 $ go install github.com/xo/usql@master
 ```
 
-Support for additional databases can be specified with [build tags][Database Support]:
-
-```sh
-# install usql with most drivers (excludes unsupported drivers)
-$ go install -tags most github.com/xo/usql
-
-# install usql with all drivers
-$ go install -tags all github.com/xo/usql
-```
-
 ## Building
 
 When building `usql` with [Go][go-project], only drivers for PostgreSQL, MySQL,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -124,16 +124,6 @@ scoop install usql
...
 ```
 
-Support for additional databases can be specified with [build tags][Database Support]:
-
-```sh
-# install usql with most drivers (excludes unsupported drivers)
-$ go install -tags most github.com/xo/usql
-
-# install usql with all drivers
-$ go install -tags all github.com/xo/usql
```
jpy-git, line 134: Duplicated by lines 147 and 150 so removing.

PR #312, opened 2022-01-17
fix empty variables not shown in '\raw'

### LINKED ISSUE #257: empty variables not shown in `\raw`
Empty variables should show in `\raw`.

For example, currently `usql` produces the following output:

```sh
(not connected)=> \set a ''
(not connected)=> select :a
(not connected)-> \p
select 
(not connected)-> \raw
select 
(not connected)->
```

However, on the `\raw` above, it should have shown:

```sh
(not connected)-> \raw
select :a
(not connected)->
```

Comparatively:

```sh
(not connected)=> \set a b
(not connected)=> select :a
(not connected)-> \p
select b
(not connected)-> \raw
select :a
(not connected)->
```

And, it behaves as expected for quoted variables:

```sh
(not connected)=> \set a ''
(not connected)=> select :'a'
(not connected)-> \p
select ''
(not connected)-> \raw
select :'a'
(not connected)->
```

It's worth noting this is not dependent on the length of the variable name:

```sh
(not connected)=> select :aa
(not connected)-> \p
select 
(not connected)-> \raw
select 
(not connected)-> 
```

The reason for this behavior is partly due to the original way that variables were parsed and interpolated (at time of entry). This will require a slight overhaul to how variables are captured and interpolated.

xwjdsh:
Close #257

### DIFF
--- a/stmt/stmt.go
+++ b/stmt/stmt.go
@@ -22,8 +22,8 @@ type Var struct {
 	Name string
 	// Len is the length of the replaced variable.
 	Len int
-	// Exist indicates whether the variable has been defined.
-	Exist bool
+	// Defined indicates whether the variable has been defined.
+	Defined bool
 }
 
 // String satisfies the fmt.Stringer interface.
@@ -101,7 +101,7 @@ func (b *Stmt) RawString() string {
 	var i int
 	// deinterpolate vars
 	for _, v := range b.Vars {
-		if !v.Exist {
+		if !v.Defined {
 			continue
 		}
 		if len(s) > i {
@@ -246,7 +246,7 @@ parse:
 				}
 				b.Vars = append(b.Vars, v)
 				if ok, z, _ := unquote(q+v.Name+q, true); ok {
-					v.Exist = true
+					v.Defined = true
 					b.r, b.rlen = substituteVar(b.r, v, z)
 					i--
 				}
--- a/stmt/stmt.go
+++ b/stmt/stmt.go
@@ -22,6 +22,8 @@ type Var struct {
 	Name string
 	// Len is the length of the replaced variable.
 	Len int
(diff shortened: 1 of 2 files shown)


stmt/stmt.go
```diff
@@ -22,6 +22,8 @@ type Var struct {
 	Name string
 	// Len is the length of the replaced variable.
 	Len int
+	// Exist indicates whether the variable has been defined.
+	Exist bool
```
nineinchnick, line 26: `Exist` is a bit generic, how about `IsDefined`?
```suggestion
	IsDefined bool
```

kenshaw: `Defined` ...

stmt/stmt_test.go
```diff
@@ -254,6 +254,29 @@ func TestNextResetState(t *testing.T) {
 	}
 }
 
+func TestEmptyVariablesRawString(t *testing.T) {
+	stmt := new(Stmt)
+	stmt.Buf = []rune("select ")
```
kenshaw, line 259: A unit test should not be setting this manually. Use `Append`, which is what's used internally on the call back from `rline`.

stmt/stmt_test.go
```diff
@@ -254,6 +254,29 @@ func TestNextResetState(t *testing.T) {
...
+	stmt.Prefix = "SELECT"
+	v := &Var{
+		I:    7,
+		End:  9,
+		Name: "a",
+		Len:  0,
+	}
+	stmt.Vars = append(stmt.Vars, v)
+
+	if expected, got := "select ", stmt.RawString(); expected != got {
```
kenshaw, line 270: Please change `expected` -> `exp`.

xwjdsh:
I udpated.

kenshaw:
Added.

PR #335, opened 2022-06-13
Do not use pager when \watch

### LINKED ISSUE #313: \watch should not use PAGER
Whenever a pager is used in \watch, the PAGER needs to be terminated manually.

If \watch wants to use the PAGER, `usql` will have to output to the pager (without exiting).
See PSQL_WATCH_PAGER in PostgreSQL15.

noborus:
Users do not want the pager to start every time it is run with \watch.
Resolve #313


kenshaw:
Added.

Issue #177, opened 2021-02-06
\d commands with a data type column should include scale and precision


nineinchnick:
\d commands should print data types with scale and precision in parenthesis (like `varchar(255)` or `decimal(4,2)`), not as separate columns, to make it easier to read by humans.

PR #343, opened 2022-07-14
Add dataTypeFormatter to information schema reader

henlue:
Implements  #177

Adds a `dataTypeFormatter` function to the `InformationSchema` reader. The function builds the datatype string from a `metadata.Column` object.

Only working for the postgres driver so far. For the interval type the fields (second, year ...) are not printed. But this is unrelated to this PR.

```
|         usql Type              |         psql Type              |
+--------------------------------+--------------------------------+
| bit(1)                         | bit(1)                         |
| bit varying(2)                 | bit varying(2)                 |
| character(3)                   | character(3)                   |
| character varying(4)           | character varying(4)           |
| interval                       | interval second(6)             |
| numeric(1,0)                   | numeric(1,0)                   |
| time(4) without time zone      | time(4) without time zone      |
| time(3) with time zone         | time(3) with time zone         |
| timestamp(2) without time zone | timestamp(2) without time zone |
| timestamp(1) with time zone    | timestamp(1) with time zone    |
| bit(1)                         | bit(1)                         |
| bit varying                    | bit varying                    |
| character(1)                   | character(1)                   |
| character varying              | character varying              |
| interval                       | interval                       |
| interval                       | interval second                |
| interval                       | interval year                  |
| numeric                        | numeric                        |
| time(6) without time zone      | time without time zone         |
| time(6) with time zone         | time with time zone            |
| timestamp(6) without time zone | timestamp without time zone    |
| timestamp(6) with time zone    | timestamp with time zone       |
| bigint                         | bigint                         |
| bigint                         | bigint                         |
| boolean                        | boolean                        |
| box                            | box                            |
| bytea                          | bytea                          |
| cidr                           | cidr                           |
| circle                         | circle                         |
| date                           | date                           |
| double precision               | double precision               |
| inet                           | inet                           |
| integer                        | integer                        |
| json                           | json                           |
| jsonb                          | jsonb                          |
| line                           | line                           |
| lseg                           | lseg                           |
| macaddr                        | macaddr                        |
| macaddr8                       | macaddr8                       |
| money                          | money                          |
| path                           | path                           |
| pg_lsn                         | pg_lsn                         |
| pg_snapshot                    | pg_snapshot                    |
| point                          | point                          |
| polygon                        | polygon                        |
| real                           | real                           |
| smallint                       | smallint                       |
| smallint                       | smallint                       |
| integer                        | integer                        |
| text                           | text                           |
| tsvector                       | tsvector                       |
| txid_snapshot                  | txid_snapshot                  |
| uuid                           | uuid                           |
| xml                            | xml                            |
```


drivers/metadata/postgres/metadata.go
```diff
@@ -34,7 +34,44 @@ func NewReader() func(drivers.DB, ...metadata.ReaderOption) metadata.Reader {
 			}),
 			infos.WithSystemSchemas([]string{"pg_catalog", "pg_toast", "information_schema"}),
 			infos.WithCurrentSchema("CURRENT_SCHEMA"),
-		)
+			infos.WithDataTypeFormatter(func(col metadata.Column) string {
```
nineinchnick, line 37: Can you define this as a standalone function and reference it here? It's a bit big for a lambda.

drivers/metadata/informationschema/metadata.go
```diff
@@ -158,6 +160,14 @@ func WithCurrentSchema(expr string) metadata.ReaderOption {
 	}
 }
 
+// WithDataTypeFormatter function to build updated string represenation of data type
+// from Column
+func WithDataTypeFormatter(form func(metadata.Column) string) metadata.ReaderOption {
```
nineinchnick, line 165: Nit: replace `form` with either `f` or `formatter`, `form` is a bit misleading as it is a regular word with a different meaning.

drivers/metadata/postgres/metadata.go
```diff
@@ -34,7 +34,44 @@ func NewReader() func(drivers.DB, ...metadata.ReaderOption) metadata.Reader {
...
+					}
+				case "character":
+					return fmt.Sprintf("character(%d)", col.ColumnSize)
+				case "character varying":
+					if col.ColumnSize != 0 {
+						return fmt.Sprintf("character varying(%d)", col.ColumnSize)
+					} else {
+						return col.DataType
+					}
+				// case "interval":
```
nineinchnick, line 55: Remove commented out code.

nineinchnick (commented):
Can you also add a test? We only had a test for column names, so either add a new one or expand the existing one to check other fields too.

henlue:
Data type formatting is now implemented for postgres, mysql and sqlserver.

I've also included the ReadinessURL fix for the sqlserver tests. On my machine the tests now at least work if I include `--cleanup=false`. Without the flag they still fail though. The problem is that while the code checks if the sakila schema is there, it is not checking whether all tables have actually been created already.

To fix this properly I would propose to execute the commands from `sql-server-sakila-schema.sql` directly from the go testing code and not through a separate shell command. Like this it will be obvious to the testing code when the database initialization has finished, and it would reduce complexity, because building docker images for the tests would not be necessary anymore.

Anyways, since this would be a bigger change, I would propose to do this as a separate PR and close this PR in the current state.

drivers/metadata/mysql/metadata.go
```diff
@@ -17,6 +17,7 @@ var (
 		infos.WithSequences(false),
 		infos.WithCheckConstraints(false),
 		infos.WithCustomClauses(map[infos.ClauseName]string{
+			infos.ColumnsDataType:                 "column_type",
```
nineinchnick, line 20: I wonder how this worked with `data_type`, or maybe it didn't?

nineinchnick (commented):
Excellent work, thanks!

To fix sqlserver tests, I guess we need  a readiness probe that'd wait for a table to exists, but you're definitely right we should do this in a separate PR.

drivers/metadata/mysql/metadata.go
henlue: It did work with `data_type`, but in mysql there is an additional `column_type` column that already contains the datatype in the format we are looking for.

PR #349, opened 2022-07-30
Release usql with goreleaser & Github Actions

anthonyroussel:
Hello

I saw that you don't have a pipeline for releasing usql automatically and on several platforms (linux, darwin, windows... + amd64/arm64).

I suggest you use [GoReleaser](https://github.com/goreleaser/goreleaser-cross) & [GoReleaser Cross](https://github.com/goreleaser/goreleaser-cross) to solve this problem.

With the Github Action pipeline, this will allow to build and create a Github release automatically when pushing a new Git tag.

What do you think about this?


anthonyroussel:
Tested on my fork repository: https://github.com/anthonyroussel/usql/actions/runs/2767199327

kenshaw:
@anthonyroussel This has been proposed before. I'm not _against_ the idea, but I'm opposed to fully automating releases  completely. Is there a way to download the binaries generated from that push? I don't see any on that GitHub project/repo. I'd like to look at/verify/test the binaries.

kenshaw:
@anthonyroussel I was able to locate and download the binaries. As expected, the binaries for Windows and macOS do not work. I tried both the `arm64` and `amd64` binaries on macOS. If you can figure out a way to overcome this, I'd be interested in doing a more automated release process. Cheers.

anthonyroussel:
Oops, indeed I don't have any simple way to check the MacOS binaries.

I instanciated a MacOS's environment on Github Action + [SSH](https://github.com/marketplace/actions/debugging-with-tmate) and looks like the `Killed: 9` error for the MacOS binary is caused by a bug in the Debian's upx package: https://github.com/upx/upx/issues/424

:/

kenshaw:
You can't use `upx` cross platform. GitHub actions have the ability to run/target `macos`. Can you try it with that?

anthonyroussel:
> You can't use `upx` cross platform. GitHub actions have the ability to run/target `macos`. Can you try it with that?

Hello

Yes indeed, you're right!

It's more reliable to build `usql` directly on the target distribution rather than to cross-compile for all platforms from the Ubuntu runner.

I made a working version with this pipeline https://github.com/anthonyroussel/usql/commit/release-pipeline

This pipeline contains 4 Github Action jobs:

* One job to build the linux amd64+amd64 static binaries from the Linux runner
* Another one to build the macOS amd64 binary from the MacOS runner
* Another distinct macOS build job to build the macOS arm64 binary (to avoid conflicts with the MacOS amd64 jobs and compiled CGO dependencies) from the MacOS runner
* And a last to build the Windows amd64 binary from the Windows runner

Looks like the windows and linux binaries are working.
Unfortunately I can't test the MacOS builds (amd64 + arm64) very easily.

The generated test release is available here: https://github.com/anthonyroussel/usql/releases/tag/v0.12.0

What do you think about this version? :)

kenshaw:
@anthonyroussel I'll test the binaries. If I had to guess, the `arm64` Darwin versions won't work, but hey, if we can get it automated for the other platforms, all the better.

kenshaw:
I've merged your changes. The `darwin/arm64` binary doesn't work, but I believe just turning off `upx` will make it work. I've modified the `build-release.sh` and I just tagged a release. Once it's done building, I'll do another round of checks on the built binaries, hopefully this works.

kenshaw:
@anthonyroussel Really appreciate your work here. Unfortunately, the `darwin/arm64` builds don't work (with or without `upx`). Will just manually build for that platform/arch moving forward, until there is a way to do it in an automated fashion. Cheers.

sumanthkumarc:
FWIW, `darwin/amd64` builds aren't working too. have been trying to test the binaries on my intel mac. It always gives `zsh: exec format error: ./usql` wierd because i did select correct platform and arch.

PR #352, opened 2022-08-03
fix: Resolves #351

### LINKED ISSUE #351: Shows columns from tables with same name in different database in clickhouse
I have two tables with same name (`tab`) in different databases (`one` and `two`). 
```
ch:localhost=> \dt+ tab
               List of relations
 Schema | Name | Type  | Rows | Size | Comment
--------+------+-------+------+------+---------
 one    | tab  | TABLE |    0 | 0    |
 two    | tab  | TABLE |    0 | 0    |
```
Both tables have one column each.
```
ch:localhost=> select * from one.tab;
 ok
----
 1
(1 row)

ch:localhost=> select * from two.tab;
 notok
-------
 1
(1 row)

```
But, if I use `\t one.tab`, it show columns from both the tables.
```
ch:localhost=> \d one.tab
           TABLE "one.tab"
 Name  |  Type  | Nullable | Default
-------+--------+----------+---------
 notok | String | ""       |
 ok    | String | ""       |
```

v4run:


drivers/clickhouse/clickhouse_test.go
```diff
@@ -116,10 +116,10 @@ func TestColumns(t *testing.T) {
 	if err != nil {
 		log.Fatalf("could not read columns: %v", err)
 	}
-	checkNames(t, "column", res, colNames()...)
+	checkNames(t, "column", true, res, colNames()...)
 }
 
-func checkNames(t *testing.T, typ string, res interface{ Next() bool }, exp ...string) {
+func checkNames(t *testing.T, typ string, checkForAdditional bool, res interface{ Next() bool }, exp ...string) {
```
nineinchnick, line 122: I'm not sure `checkForAdditional` is necessary, can we always check?

v4run: I was thinking about the case where new functions gets added in the newer versions of clickhouse.

nineinchnick: Then we should pin the version of CH we use for testing to a particular one and fix tests whenever we wtorek update it.

v4run: Yeah. That is a good idea. I will make the necessary changes.

nineinchnick (approved):

Issue #172, opened 2021-02-06
Implement privilege related \d commands


nineinchnick:
Privilege related \d commands include:
* [x] \dp     [PATTERN]      list table, view, and sequence access privileges - with `\z` as an alias
* [ ] \ddp    [PATTERN]      list default privileges
* [ ] \dg[S+] [PATTERN]      list roles
* [ ] \du[S+] [PATTERN]      list roles
* [ ] \drds [PATRN1 [PATRN2]] list per-database role settings

PR #354, opened 2022-08-09
Add /dp command to information_schema reader

henlue:
Adds the /dp backslash command. It lists privileges that have been granted on tables, views and sequences.

It is implemented in the information schema metadata reader and was tested against postgres, sqlserver and mariadb. I could not test it against snowflake and netezza, since I don't have access to a development instance.

Partly implements #172

### DIFF
--- a/drivers/metadata/informationschema/metadata.go
+++ b/drivers/metadata/informationschema/metadata.go
@@ -22,6 +22,9 @@ type InformationSchema struct {
 	hasIndexes          bool
 	hasConstraints      bool
 	hasCheckConstraints bool
+	hasTablePrivileges  bool
+	hasColumnPrivileges bool
+	hasUsagePrivileges  bool
 	clauses             map[ClauseName]string
 	limit               int
 	systemSchemas       []string
@@ -56,6 +59,8 @@ const (
 	ConstraintJoinCond          = ClauseName("constraint_join.fk")
 
 	SequenceColumnsIncrement = ClauseName("sequence_columns.increment")
+
+	PrivilegesGrantor = ClauseName("privileges.grantor")
 )
 
 // New InformationSchema reader
@@ -67,6 +72,9 @@ func New(opts ...metadata.ReaderOption) func(drivers.DB, ...metadata.ReaderOptio
 		hasIndexes:          true,
 		hasConstraints:      true,
 		hasCheckConstraints: true,
+		hasTablePrivileges:  true,
+		hasColumnPrivileges: true,
+		hasUsagePrivileges:  true,
 		clauses: map[ClauseName]string{
 			ColumnsDataType:                 "data_type",
 			ColumnsColumnSize:               "COALESCE(character_maximum_length, numeric_precision, datetime_precision, 0)",
@@ -81,6 +89,7 @@ func New(opts ...metadata.ReaderOption) func(drivers.DB, ...metadata.ReaderOptio
 			ConstraintIsDeferrable:          "t.is_deferrable",
 			ConstraintInitiallyDeferred:     "t.initially_deferred",
 			SequenceColumnsIncrement:        "increment",
+			PrivilegesGrantor:               "grantor",
 		},
 		systemSchemas:     []string{"information_schema"},
 		dataTypeFormatter: func(col metadata.Column) string { return col.DataType },
@@ -145,6 +154,27 @@ func WithCheckConstraints(con bool) metadata.ReaderOption {
 func WithSequences(seq bool) metadata.ReaderOption {
 	return func(r metadata.Reader) {
 		r.(*InformationSchema).hasSequences = seq
+	}
+}
+
+// WithTablePrivileges when the `table_privileges` table exists
+func WithTablePrivileges(t bool) metadata.ReaderOption {
+	return func(r metadata.Reader) {
+		r.(*InformationSchema).hasTablePrivileges = t
+	}
+}
+
+// WithColumnPrivileges when the `column_privileges` table exists
+func WithColumnPrivileges(c bool) metadata.ReaderOption {
+	return func(r metadata.Reader) {
+		r.(*InformationSchema).hasColumnPrivileges = c
+	}
+}
+
+// WithUsagePrivileges when the `usage_privileges` table exists
+func WithUsagePrivileges(u bool) metadata.ReaderOption {
+	return func(r metadata.Reader) {
+		r.(*InformationSchema).hasUsagePrivileges = u
 	}
 }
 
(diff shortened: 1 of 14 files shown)


drivers/metadata/informationschema/metadata.go
```diff
@@ -821,6 +851,179 @@ func (s InformationSchema) Sequences(f metadata.Filter) (*metadata.SequenceSet,
...
+	}
+
+	// In the query result, table and column level privileges will be on separate rows.
+	// Each table or column can have multple privileges (i.e rows).
+	// For table level privileges the `column_name` column is empty.
+	qstr := "SELECT * FROM (\n" + strings.Join(qstrs, "\nUNION ALL\n") + "\n) AS subquery"
+	rows, closeRows, err := s.query(
+		qstr,
+		conds,
+		"object_catalog, object_schema, object_name, object_type, column_name, grantee, grantor, privilege_type",
```
nineinchnick, line 938: Should results be sorted by object type first, and then object name?

drivers/metadata/informationschema/metadata.go
```diff
@@ -821,6 +851,179 @@ func (s InformationSchema) Sequences(f metadata.Filter) (*metadata.SequenceSet,
...
+		r := row{}
+		err = rows.Scan(&r.Catalog, &r.Schema, &r.Name, &r.ObjectType, &r.Column, &r.Grantee, &r.Grantor, &r.PrivilegeType, &r.IsGrantable)
+		if err != nil {
+			return nil, err
+		}
+
+		switch {
+		// If the row specifies neither column nor table level privileges
+		case r.PrivilegeType == "":
+			if curSummary.Catalog != r.Catalog || curSummary.Schema != r.Schema || curSummary.Name != r.Name {
```
nineinchnick, line 973: Reverse the condition and do a `continue` to avoid one level of indentation.

drivers/metadata/metadata.go
```diff
@@ -831,6 +843,175 @@ func (s Sequence) values() []interface{} {
...
+	}
+	return p[i].PrivilegeType < p[j].PrivilegeType
+}
+
+func (p ObjectPrivileges) String() string {
+	if len(p) == 0 {
+		return ""
+	}
+
+	sort.Sort(p)
```
nineinchnick, line 934: Can this be done once on struct initalization?

nineinchnick (commented):
Great work! I did the first pass, will review more later. Can you add examples how the results tables look like in different databases?

henlue:
Some examples:

Sqlserver
```
                                      Access privileges
 Schema |         Name          |    Type    |  Access privileges   |    Column privileges    
--------+-----------------------+------------+----------------------+-------------------------
 dbo    | MSreplication_options | BASE TABLE |                      |  
 dbo    | spt_fallback_db       | BASE TABLE | public=SELECT/dbo    |  
 dbo    | spt_fallback_dev      | BASE TABLE | public=SELECT/dbo    |  
 dbo    | spt_fallback_usg      | BASE TABLE | public=SELECT/dbo    |  
 dbo    | spt_monitor           | BASE TABLE | public=SELECT/dbo    |  
 dbo    | spt_values            | VIEW       | public=SELECT/dbo    |  
 dbo    | test_table            | BASE TABLE | test_user=SELECT/dbo | id:                    +
        |                       |            |                      |   test_user=UPDATE*/dbo
```

Maria DB
```
                                       Access privileges
    Schema     |    Name    |    Type    |    Access privileges    |     Column privileges     
---------------+------------+------------+-------------------------+---------------------------
 test_database | test_table | BASE TABLE | 'test_user'@'%'=SELECT* | id:                      +
               |            |            |                         |   'test_user'@'%'=INSERT* 
(1 row)
```

Postgres
```
                                                                               Access privileges
 Schema |    Name    |    Type    |                                Access privileges                                 |                    Column privileges                    
--------+------------+------------+----------------------------------------------------------------------------------+---------------------------------------------------------
 public | test_table | BASE TABLE | postgres=DELETE*,INSERT*,REFERENCES*,SELECT*,TRIGGER*,TRUNCATE*,UPDATE*/postgres+| id:                                                    +
        |            |            | test_user=SELECT/postgres                                                        |   postgres=INSERT*,REFERENCES*,SELECT*,UPDATE*/postgres+
        |            |            |                                                                                  |   test_user=INSERT*,SELECT/postgres                    +
        |            |            |                                                                                  | name:                                                  +
        |            |            |                                                                                  |   postgres=INSERT*,REFERENCES*,SELECT*,UPDATE*/postgres+
        |            |            |                                                                                  |   test_user=SELECT/postgres 
(1 row)
```

The output for postgres is quite verbose. The problem is that the owner by default has all privileges and all those privileges will be printed for each table and for each column separately. In psql the default privileges for the owner are hidden. But from the information_schema it is not possible to infer the owner of a table.

Maria DB and Sqlserver don't have this problem because they don't write the default privileges to the `information_schema` in the first place.

In psql the privileges are abreviated (e.g. `postgres=arwdDxt/postgres`). I left them in a written out form. Otherwise we would need to come up with abreviations for all privileges that are used in all supported databases. But this also makes the output longer.

I guess a workaround to help keep the overview for bigger schemas might be to only show the "Column priviliges" column with `\dp+`. A better solution would be a postgres specific implentation that has access to the table owner and can filter the output based on that.

nineinchnick:
Thanks, these examples look neat!

> I guess a workaround to help keep the overview for bigger schemas might be to only show the "Column priviliges" column with `\dp+`. A better solution would be a postgres specific implentation that has access to the table owner and can filter the output based on that.

When I was implementing other metadata items earlier I ended up having to use PostgreSQL's system catalogs anyway, instead of the information schema.

Does it make sense to cram both table and column privileges into a single table? Maybe it'll be easier to see if the `PrivilegeSummary` abstraction is right if there's more than one implementation, so that's another reason to try to use PostgreSQL's specific catalogs.

drivers/metadata/informationschema/metadata.go
henlue: To me that would make sense.

The information schema implementation for \d sorts by object type too. But psql and our postgres implementation of \d sort by object name. I would propose to change the \d postgres implementation to object type as well to make it consistent.

drivers/metadata/informationschema/metadata.go
henlue: Not sure if I understood that. But I made it quite a bit simpler. Just without reversing the condition.

drivers/metadata/metadata.go
henlue: Right now the privileges objects are extended iteratively. So on object initialization the slice is not yet complete. I wanted to make it easy to build the privileges objects from the query result. If you would have to collect all privileges first into a slice and then initialize the privileges object (which is just a slice anyways currently) that would make the `PrivilegeSummaries` method slightly more complex. And in practice the String method and with it the Sort would anyways only be executed once.

So I guess the current solution favors readability and less complexity over performance. But it could be changed of course.

henlue: Sure. I removed the sort. In case of the information schema reader the values are anyways sorted by the query already. Other callers could simply sort the privileges themselves. I left comments to make sure this assumption is not missed.

drivers/metadata/informationschema/metadata.go
henlue: I made the changes for both the postgres reader and the information schema reader.

henlue:
Good question. I didn't really think about it. Just copied what psql does. I guess it can be nice to have all the privileges for one db object in a single place.

So we leave it as it is for now, and implement the postgres reader in a seperate PR?

nineinchnick:
> Good question. I didn't really think about it. Just copied what psql does. I guess it can be nice to have all the privileges for one db object in a single place.
> 
> So we leave it as it is for now, and implement the postgres reader in a seperate PR?

Yes, sounds like a good plan!

drivers/metadata/metadata.go
nineinchnick: Thanks!

nineinchnick (approved):

nineinchnick:
@henlue is it ready to be merged?

henlue:
Yes :-)

PR #355, opened 2022-08-16
Fix TestWriter

henlue:
Fix API calls to include URL
Add functionality to ignore randomly named indices in test function output
Update expected test function result files
Fix small bug in the ordering of postgres indices

pgsql.descTable still fails but this is due to an actual bug in the postgres metadata reader.

I had to execute the tests with `cleanup=false`. Otherwise the db initialization sometimes does not finish in time. This issue is not yet fixed.


drivers/drivers_test.go
```diff
@@ -456,19 +457,31 @@ func TestCopy(t *testing.T) {
 	}
 }
 
-func filesEqual(a, b string) error {
+// FilesEqual compares the files at path a and b and returns an error if
```
nineinchnick, line 460:
```suggestion
// filesEqual compares the files at paths a and b and returns an error if
```

nineinchnick (approved):

drivers/drivers_test.go
henlue: done

Issue #359, opened 2022-08-23
Feature: Column metadata harmonized with row result metadata


Allam76:
`database/sql` forces drivers to use the [ColumnType](https://pkg.go.dev/database/sql#ColumnType) for query return metadata.
This is also used internally in usql.

Would it be possible to ask the column metadata from usql to also be able to conform to this format? Or at lease an extension point so that it can be done in userland? JDBC, ODBC, etc already have this mapping.

The drivers are forced to implement this conversion so if should not be too difficult. I could possibly volunteer:-).

This is useful when migrating from one database to another and for external tools. Also I understand this might not be a priority for the usql tool itself.

kenshaw:
I'm not sure what the request is here? For what it's worth, how Go "sees" the types of those columns isn't what you want `usql` reporting on the metadata queries. You want the information as the database sees it. It's worth mentioning, as well, that the `ColumnType` you're referring to is only available after a query has been executed. This isn't information that is "queryable" from the perspective of `usql`.

Allam76:
Consider JDBC for example. There the metadata is mapped to an external [JDBC type system](https://docs.oracle.com/javase/8/docs/api/java/sql/JDBCType.html) and then all client tools can use that rather than that each client tool must map to all possible databases.

I realize this might be a bit out of scope for a pure terminal-based front-end like usql. Unless if you want to add a data migration tool.

The `ColumnType` is as far as I know the only equivalent in go to the JDBC Type.

The implementation at least for postgres is rather trivial:
https://github.com/lib/pq/blob/8c6de565f76fb5cd40a5c1b8ce583fbc3ba1bd0e/rows.go#L24-L42

kenshaw:
If you want to extract metadata about tables or queries from a SQL database, please use my `xo` tool. `usql` is meant to be a generic command-line client, and is a humble reimplementation of `psql`. While you might be able to use `usql` for something like a database migration, that's really not the design/intent of `usql`, and such tasks are best accomplished with individual database's provided toolsets and/or use more specific Go packages/tools that try to provide this kind of functionality.

Also, I'm still not sure what the feature request is? If it's easier to explain with code, we're always open to PRs!

Allam76:
I'm working on a sql federated query engine. That is, a database without storage engine. So one needs to connect to other databases to fetch the data. Naturally many different databases should be supported so to stay sane some abstraction is needed.

usql seemed to have the part of the metadata: https://github.com/golang/go/issues/7408#issuecomment-829333951

[...]

Issue #358, opened 2022-08-23
Bug: postgres metadata reader; options are assigned twice


Allam76:
Consider the following code:
```go
func TestUsqlPostgresOptionsFail(t *testing.T) {
	db, _ := sql.Open("postgres", "postgres://martin:<pass>@localhost:5432/northwind")
	reader := postgres.NewReader()(db,infos.WithDataTypeFormatter(dataTypeFormatter)).(metadata.BasicReader)
	tableSet, err := reader.Tables(metadata.Filter{Schema: "public", Catalog: "northwind"})
	assert.NoError(t, err)
	assert.NotEmpty(t, tableSet)
}
```
Here I would like to add a custom data type formatter to the postgres metadata reader.
This fails with:
```panic: interface conversion: metadata.Reader is *metadata.LoggingReader, not *informationschema.InformationSchema```

From reading the offending code:
```go
func NewReader() func(drivers.DB, ...metadata.ReaderOption) metadata.Reader {
	return func(db drivers.DB, opts ...metadata.ReaderOption) metadata.Reader {
		newIS := infos.New(
			infos.WithIndexes(false),
			infos.WithCustomClauses(map[infos.ClauseName]string{
				infos.ColumnsColumnSize:         "COALESCE(character_maximum_length, numeric_precision, datetime_precision, interval_precision, 0)",
				infos.FunctionColumnsColumnSize: "COALESCE(character_maximum_length, numeric_precision, datetime_precision, interval_precision, 0)",
			}),
			infos.WithSystemSchemas([]string{"pg_catalog", "pg_toast", "information_schema"}),
			infos.WithCurrentSchema("CURRENT_SCHEMA"),
			infos.WithDataTypeFormatter(dataTypeFormatter))
		return metadata.NewPluginReader(
			newIS(db, opts...),
			&metaReader{
				LoggingReader: metadata.NewLoggingReader(db, opts...),
			},
		)
	}
}
```
it is clear that the ```opts...``` are assigned twice. Last time to an incompatible reader (LoggingReader).

Maybe also consider putting
```go
type metaReader struct
```
in uppercase to facilitate extension.

kenshaw:
@Allam76 PRs are always welcome!

nineinchnick:
[...]

PR #360, opened 2022-08-25
Discussion: Metadata; internal to external data type conversion

Allam76:
Following out discussion in #359 and #358 here comes a possible implementation.
Note this is not intended to be a ready fix but rather an idea on what I have in mind.

I've added 2 new fields to the metadata.Column. Internal and external data type.
External is the one that is well, external and should be uniform between databases. The internal comes from the fact that most databases allow to write the same data type in many different format. VCHAR, VARCHAR, VARIABLE CHARACTER, etc all are the same type. This is the underlying type to reduce the number entries in the mapping.

It is implemented for only two databases (postgres and sqlite) that covers a bit of the spectrum of available databases.

I'm also agreeing that the metadata maybe should be extracted into a more general purpose package.


nineinchnick:
What is the purpose of having unified types? Would it be used only for display in the describe commands (`\d`) or for formatting values when presenting query results? This is kind of a slippery slope, because you'd have to be careful when mapping types to make sure they match, that is any conversion is not lossy. This is true even when only displaying them to users.

If there are equivalent types in a database, like PostgreSQL, maybe consider normalizing them inside a single metadata reader?

Allam76:
I have two use cases:
1) Federated SQL engine. This is an engine without own storage. So it uses other (several) databases underneath. Like one database to rule them all. Presto/Trino is an example in the enterprise sphere. This engine expects a metadata format and writing one mapping per individual database is tedious.

2) Migration. I have an ancient database with 14 K tables to migrate with metadata in yet another database so total databases to handle is three. Then having unified metadata is a great plus.

I'll also have a look at your xo for the migration. This could potentially be helpful.

nineinchnick:
I'm a Trino contributor and I work for Starburst Data. I wrote a few Trino connectors so I have hands on experience with data type mapping. 

You have not answered my questions why unifying types helps in those use cases. Right now it looks like you only need to read types. My concern is that you can't simply define type aliases, without also defining how to convert data to avoid loosing precision.

Allam76:
Aha ok, I understand your concern now. I was simple assuming from my exploration that the rest of the data needed to avoid loosing precision is available as is.

If not, then indeed I have a bigger problem.

nineinchnick:
Maybe we could report the Go's scan type, but I have a very subjective feeling most drivers don't implement this at all. Since there are very few Go types, I'm not sure if that'll be useful.

PR #365, opened 2022-09-03
Update go-mssqldb driver reference

burningalchemist:
Notable changes:
- Replace `denisenkom/go-mssqldb` to a maintained compatible `microsoft/go-mssqldb` driver;
- Add `azuresql` alias.

With this change we only replace the driver and add `azuresql` alias for compatibility with `xo` and `dburl` modules.

To connect to MSSQL DB via Azure AD we need to import the submodule `azuread`. We could potentially replace `microsoft/go-mssqldb` with `microsoft/go-mssqldb/azuread` as it implicitly imports the root module and doesn't seem to have any bad implications. However, I'm not sure if it's a good idea, or we'd better treat `azuresql` differently.

It also might be a separate PR to keep the scope small.

References:
- https://github.com/xo/dburl/pull/21


nineinchnick:
Looks like this has been merged in c39cb6c2b431dcc2f223ff16efa130212c438241

Issue #372, opened 2022-09-12
Missing auto complete when using TABLE command


kozikowskik:
There is no tables name auto-completion after hitting the `TAB` key when using the `TABLE` command.

PR #373, opened 2022-09-12
Adding TABLE command to completer.

kozikowskik:
Fix for #372. Adding missing TABLE command to completer. These changes fix the lack of `TAB` auto-completion.

### DIFF
--- a/drivers/completer/completer.go
+++ b/drivers/completer/completer.go
@@ -435,6 +435,10 @@ func (c completer) complete(previousWords []string, text []rune) [][]rune {
 	if TailMatches(IGNORE_CASE, previousWords, "FROM|JOIN") {
 		return c.completeWithSelectables(text)
 	}
+	/* TABLE, but not TABLE embedded in other commands */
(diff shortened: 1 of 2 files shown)


drivers/completer/completer.go
```diff
@@ -435,6 +435,10 @@ func (c completer) complete(previousWords []string, text []rune) [][]rune {
 	if TailMatches(IGNORE_CASE, previousWords, "FROM|JOIN") {
 		return c.completeWithSelectables(text)
 	}
+	/* TABLE <table> */
+	if TailMatches(IGNORE_CASE, previousWords, "TABLE") {
```
nineinchnick, line 439:
```suggestion
	/* TABLE, but not TABLE embedded in other commands */
	if matches(IGNORE_CASE, previousWords, "TABLE") {
```
See how it's done in `psql`: https://github.com/postgres/postgres/blob/master/src/bin/psql/tab-complete.c#L4384

kozikowskik: Fixed.

PR #376, opened 2022-09-22
Use scan types reported by driver when scanning values for copy

nineinchnick:
This should be more correct than relying on a driver to properly handle an empty interface. I checked how the type is used in `go/1.19.1/libexec/src/database/sql/sql_test.go`.


kenshaw:
@nineinchnick Is this at least of equal performance? I had originally / purposefully avoided using `reflect` due to performance concerns.

nineinchnick:
It's done only once when executing the query. I can write a benchmark for copy, we sure would use some.

PR #378, opened 2022-11-08
Add ON_ERROR_STOP: halt execution when a statement fails

sfc-gh-pbennes:
This PR adds a variable, `ON_ERROR_STOP`, disabled by default to maintain status quo, that allows usql to stop execution in the case where it's being requested to execute multiple statements, either by file input or multiple statements on the command line.

Here are some examples:

SQL file execution with file:
```sql
select 1;
select asdfa;
select 2;
```

```sh
[pbennes@localhost usql]$ ./usql -v ON_ERROR_STOP=0 "snowflake://redacted" -f test.sql
Connected with driver snowflake (<unknown>)
 1 
---
 1 
(1 row)

error: snowflake: 904: SQL compilation error: error line 1 at position 7
invalid identifier 'ASDFA'
 2 
---
 2 
(1 row)

[pbennes@localhost usql]$ ./usql -v ON_ERROR_STOP=1 "snowflake://redacted" -f test.sql
Connected with driver snowflake (<unknown>)
 1 
---
 1 
(1 row)

error: snowflake: 904: SQL compilation error: error line 1 at position 7
invalid identifier 'ASDFA'
```

`-c`:
```
[pbennes@localhost usql]$ ./usql -v ON_ERROR_STOP=0 "snowflake://redacted" -c 'select 1; select asdf; select 2;'
Connected with driver snowflake (<unknown>)
 1 
---
 1 
(1 row)

error: snowflake: 904: SQL compilation error: error line 1 at position 7
invalid identifier 'ASDF'
 2 
---
 2 
(1 row)

[pbennes@localhost usql]$ ./usql -v ON_ERROR_STOP=1 "snowflake://redacted" -c 'select 1; select asdf; select 2;'
Connected with driver snowflake (<unknown>)
 1 
---
 1 
(1 row)

error: snowflake: 904: SQL compilation error: error line 1 at position 7
invalid identifier 'ASDF'
```

And then a garbage value for the parameter:
```sh
[pbennes@localhost usql]$ ./usql -v ON_ERROR_STOP=asdf "snowflake://redacted" -c 'select 1; select asdf; select 2;'
Connected with driver snowflake (<unknown>)
 1 
---
 1 
(1 row)

error parsing value for ON_ERROR_STOP: strconv.ParseBool: parsing "asdf": invalid syntax
error: snowflake: 904: SQL compilation error: error line 1 at position 7
invalid identifier 'ASDF'
 2 
---
 2 
(1 row)
```

Interactive mode:
```
[pbennes@localhost usql]$ ./usql "snowflake://redacted"
Connected with driver snowflake (<unknown>)
Type "help" for help.

sf:pbennes@127=> select 1; select asdf; select 2;
 1 
---
 1 
(1 row)

error: snowflake: 904: SQL compilation error: error line 1 at position 7
invalid identifier 'ASDF'
 2 
---
 2 
(1 row)

sf:pbennes@127=> \set ON_ERROR_STOP 1
sf:pbennes@127=> select 1; select asdf; select 2;
 1 
---
 1 
(1 row)

error: snowflake: 904: SQL compilation error: error line 1 at position 7
invalid identifier 'ASDF'
sf:pbennes@127=>  
```

### DIFF
--- a/env/types.go
+++ b/env/types.go
@@ -36,6 +36,15 @@ func (v Vars) All() map[string]string {
 
 var vars, pvars Vars
 
+// a map of functions to optionally parse parameters when being set
+// Takes: value, name string
+// Returns: parsedValue string, error
+// See ParseBool for an example
+var variableParseFunctions = map[string]func(string, string) (string, error){
+	"ON_ERROR_STOP": ParseBoolEmptyIsOn,
+	"QUIET":         ParseBoolEmptyIsOn,
+}
+
 func init() {
 	// get USQL_* variables
 	enableHostInformation := "true"
@@ -69,6 +78,7 @@ func init() {
 		"SHOW_HOST_INFORMATION": enableHostInformation,
 		"PAGER":                 pagerCmd,
 		"EDITOR":                editorCmd,
+		"ON_ERROR_STOP":         "off",
 		// prompts
 		"PROMPT1": "%S%N%m%/%R%# ",
 		// syntax highlighting variables
@@ -125,21 +135,19 @@ func ValidIdentifier(n string) error {
 }
 
 // Set sets a variable.
+// Values are optionally parsed by a helper function if one is defined in
+// variableParseFunctions
 func Set(name, value string) error {
-	err := ValidIdentifier(name)
-	if err != nil {
+	if err := ValidIdentifier(name); err != nil {
 		return err
 	}
-	switch name {
-	case "QUIET":
-		if value == "" {
-			value = "on"
-		} else {
-			value, err = ParseBool(value, name)
-			if err != nil {
-				return err
-			}
+	// parse and validate value if there's a function defined in variableParseFunctions
+	if parseFunction, ok := variableParseFunctions[name]; ok {
+		parsedValue, err := parseFunction(value, name)
+		if err != nil {
+			return fmt.Errorf(text.FormatFieldInvalid, value, name)
 		}
+		value = parsedValue
 	}
 	vars.Set(name, value)
 	return nil
@@ -210,6 +218,14 @@ func ParseBool(value, name string) (string, error) {
 		return "off", nil
 	}
 	return "", fmt.Errorf(text.FormatFieldInvalidValue, value, name, "Boolean")
+}
+
+// ParseBoolEmptyIsOn works like ParseBool but treats empty values as "on"
+func ParseBoolEmptyIsOn(value, name string) (string, error) {
+	if value == "" {
+		return "on", nil
+	}
+	return ParseBool(value, name)
 }
 
 func ParseKeywordBool(value, name string, keywords ...string) (string, error) {
--- a/handler/handler.go
+++ b/handler/handler.go
@@ -375,7 +375,18 @@ func (h *Handler) Run() error {
 				ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
 				if err = h.Execute(ctx, out, opt, h.lastPrefix, h.last, forceBatch); err != nil {
 					lastErr = WrapErr(h.last, err)
-					fmt.Fprintln(stderr, "error:", err)
+					if env.All()["ON_ERROR_STOP"] == "on" {
+						if iactive {
+							fmt.Fprintln(stderr, "error:", err)
+							h.buf.Reset([]rune{}) // empty the buffer so no other statements are run
+							continue
+						} else {
+							stop()
+							return err
+						}
+					} else {
+						fmt.Fprintln(stderr, "error:", err)
+					}
 				}
 				stop()
 			}


handler/handler.go
```diff
@@ -375,7 +375,15 @@ func (h *Handler) Run() error {
 				ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
 				if err = h.Execute(ctx, out, opt, h.lastPrefix, h.last, forceBatch); err != nil {
 					lastErr = WrapErr(h.last, err)
-					fmt.Fprintln(stderr, "error:", err)
+					ON_ERROR_STOP, parseErr := strconv.ParseBool(env.All()["ON_ERROR_STOP"])
+					if parseErr != nil {
+						fmt.Fprintln(stderr, "error parsing value for ON_ERROR_STOP:", parseErr)
+					}
+					if ON_ERROR_STOP {
+						return err
```
nineinchnick, line 383: Call `stop()` before returning

nineinchnick: What would happen if `ON_ERROR_STOP` is enabled in interactive mode?

handler/handler.go
```diff
@@ -375,7 +375,15 @@ func (h *Handler) Run() error {
 				ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
 				if err = h.Execute(ctx, out, opt, h.lastPrefix, h.last, forceBatch); err != nil {
 					lastErr = WrapErr(h.last, err)
-					fmt.Fprintln(stderr, "error:", err)
+					ON_ERROR_STOP, parseErr := strconv.ParseBool(env.All()["ON_ERROR_STOP"])
+					if parseErr != nil {
```
nineinchnick, line 379: Can this be checked before executing the query? Failing fast usually gives better UX.

handler/handler.go
sfc-gh-pbennes: Will add stop() prior to the return.

Good Q! In interactive mode any error causes the CLI to exit, which isn't consistent with current behavior and doesn't seem desirable. Will need to figure out how to get `ON_ERROR_STOP` behavior in interactive mode without causing usql to exit.

handler/handler.go
sfc-gh-pbennes: There are probably a few places ON_ERROR_STOP can be checked/parsed though I'm not super familiar with the usql code. It needs to be checked both on startup or in the case someone calls `\set` in interactive. Do you know if there's a good place to move this to offhand? I'll look as well.

nineinchnick: Not from the top of my head. I'm not sure if other boolean vars are validated or not.

handler/handler.go
sfc-gh-pbennes: Ok, both points here should be addressed now.

handler/handler.go
sfc-gh-pbennes: Ok so I added a mechanism for optional param validation functions and added a case for ON_ERROR_STOP. One thing to note is that if the value is being set on the command line to an invalid value it does not get set but the error is silent. Here's what it looks like in practice:

```sh
[pbennes@localhost usql]$ ./usql -v ON_ERROR_STOP=asdf "snowflake://redacted" 
Connected with driver snowflake (<unknown>)
Type "help" for help.

sf:admin@127=> \set
ON_ERROR_STOP = 'false'

sf:admin@127=> \set ON_ERROR_STOP 1
sf:admin@127=> \set ON_ERROR_STOP 0
sf:admin@127=> \set ON_ERROR_STOP asdf
error: invalid value asdf for var ON_ERROR_STOP
```

env/types.go
```diff
@@ -126,23 +143,33 @@ func ValidIdentifier(n string) error {
...
+	if valueIsValid, ok := variableValidationFunctions[name]; ok {
+		if !valueIsValid(value) {
+			return fmt.Errorf(text.FormatFieldInvalid, value, name)
 		}
 	}
-	vars.Set(name, value)
-	return nil
+	// QUIET accepts special values
+	// other vars are a simple set
+	if name != "QUIET" {
```
nineinchnick, line 157: I'm not sure replacing `switch` makes it more readable. Maybe there should be a map of variable value parser functions, not just validators? Then you could move the special case for `QUIET` into a dedicated function.

handler/handler.go
```diff
@@ -375,7 +375,19 @@ func (h *Handler) Run() error {
 				ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt)
 				if err = h.Execute(ctx, out, opt, h.lastPrefix, h.last, forceBatch); err != nil {
 					lastErr = WrapErr(h.last, err)
-					fmt.Fprintln(stderr, "error:", err)
+					ON_ERROR_STOP, _ := strconv.ParseBool(env.All()["ON_ERROR_STOP"])
```
nineinchnick, line 378: Do we have to parse it again, or is it guaranteed to have only specific values here, after the changes in `Set()`?

nineinchnick (commented):
Looks better, almost there!

handler/handler.go
sfc-gh-pbennes: The idea behind using strconv.ParseBool is to allow any values strconv.ParseBool allows, but that also means having to parse it to check it. Using QUIET as an example, it only allows the values in types.go's ParseBool and is checked via string comparison for use: https://github.com/xo/usql/blob/master/handler/handler.go#L948. This could be switched to also using type.go ParseBool and be a string check against "on" instead.

env/types.go
sfc-gh-pbennes: A switch for one case seemed redundant if the control flow just handles the non-QUIET-case first. Related to the other comment though ON_ERROR_STOP could be adjusted to behave consistent with QUIET which might be better.

sfc-gh-pbennes:
Ok take a look at this most recent commit. I use types.go's ParseBool for ON_ERROR_STOP as well, which makes it consistent with the set behavior of QUIET. I push the case of `\set QUIET` with the empty value being interpreted as "on" in to ParseBool. This simplifies Set() nicely.

nineinchnick (commented):
Keeps getting better, thanks for bearing with me, just one more comment and two nitpicks.

env/types.go
```diff
@@ -204,7 +213,7 @@ var (
 
 func ParseBool(value, name string) (string, error) {
 	switch strings.ToLower(value) {
-	case "1", "t", "tr", "tru", "true", "on":
+	case "1", "t", "tr", "tru", "true", "on", "":
```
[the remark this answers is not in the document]
sfc-gh-pbennes: My read of this section is that it looks like SQL session variables unrelated to client variables and using the standard libraries strconv.ParseBool vs the somewhat awkwardly named env package's [ParseBool](https://github.com/xo/usql/blob/master/env/types.go#L205)? Is there a code path through here that interacts with env.ParseBool I'm missing?

I looked at the other callers of env.ParseBool and it looked like the semantic change of "" == "on" would be ok as implied / consistent with QUIET for `\set` and `\pset` though it _is_ still a change:
- For `\timing` https://github.com/xo/usql/blob/master/metacmd/cmds.go#L418
- For boolean Pvars that control output formatting: https://github.com/xo/usql/blob/master/env/types.go#L320

I think you're right though that a semantics change should be avoided so I'll create a `ParseBoolEmptyOn` and use that as the parser func.

If someone later decided they want to change any semantics it would at least be easy to do so.

env/types.go
```diff
@@ -125,20 +135,19 @@ func ValidIdentifier(n string) error {
...
 		return err
 	}
-	switch name {
-	case "QUIET":
-		if value == "" {
-			value = "on"
+	// parse and validate value if there's a function defined in variableParseFunctions
+	if parseFunction, ok := variableParseFunctions[name]; ok {
+		if parsedValue, err := parseFunction(value, name); err != nil {
+			return fmt.Errorf(text.FormatFieldInvalid, value, name)
```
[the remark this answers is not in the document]
sfc-gh-pbennes: In this style the variable parsedValue is block scoped to the if/else so the else block is needed otherwise the variable goes out of scope and this causes a undefined variable compiler error.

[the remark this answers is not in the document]
nineinchnick: Yeah, you'd have to split up the assignment and the condition. Up to you, feel free to mark this comment as resolved.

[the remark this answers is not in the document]
sfc-gh-pbennes: I think your proposal is a bit more idiomatic. Let me change it.

env/types.go
```diff
@@ -212,6 +220,14 @@ func ParseBool(value, name string) (string, error) {
 	return "", fmt.Errorf(text.FormatFieldInvalidValue, value, name, "Boolean")
 }
 
+// The var QUIET and ON_ERROR_STOP treat an empty value as "on"
```
nineinchnick, line 223: Since you mentioned idiomatic code, it's idiomatic to start comments with the function name:
```suggestion
// ParseBoolEmptyIsOn works like ParseBool but treats empty values as "on"
```

nineinchnick (approved):
Please squash all commits and I'll merge this PR, thanks!

I'll create an issue to mention these environment variables in the README. There's no section for it right now so let's do it in another PR.

sfc-gh-pbennes:
Squashed! Was going to ask about documentation as well but you read my mind and answered haha. Thanks @nineinchnick

nineinchnick:
Docs issue: https://github.com/xo/usql/issues/379

kenshaw:
@sfc-gh-pbennes I didn't like the addition of a top level exported func in the `env` package that is not used anywhere else. I've cleaned up the logic, and fixed some other superficial errors in the latest commit. Can you please check that this logic still works as you expected? Thanks.

sfc-gh-pbennes:
@kenshaw makes sense, thanks for the FYI. Just checked and the logic still works as expected!

kenshaw:
Ok, thanks. I'll be tagging and pushing a point release shortly with this.

Issue #379, opened 2022-11-09
Document all supported environment variables in README


nineinchnick:
`usql` recognizes a few environmental variables, like:
* `PAGER`
* `EDITOR`
* `SHOW_HOST_INFORMATION`
* `QUIET`
* `ON_ERROR_STOP`

only some of them are documented. Add a new section in the README file.

sfc-gh-pbennes:
Somewhere around https://github.com/xo/usql#backslash-commands seems like a reasonable place for such a section.

Also worth noting that `\? options` and `\? variables` appear in the documentation and interactive help (`\?`) but don't appear to work (same with options):

```
(not connected)=> \? variables
...
\?: extra argument "variables" ignored
```

PR #387, opened 2022-12-19
Implement help subcommands

### LINKED ISSUE #258: expand documentation and add more to `\?` command
Current `usql` documentation  needs to be expanded to provide the same level of documentation level as `psql`. Additionally, the `\?` command should be able to look up specific help topics. Also, we should make the documentation  available via the `usql.app` website.

This will require effort in a number of ways:

1. Authoring the documentation in a form that can be embedded in both the `usql` command line application, and used for the website
2. Building the simple documentation site and related infrastructure configuration (like Docusaurus and custom cloud-init scripts)
3. Hooking in embedded documentation to the `\?` command, as per `psql`

### LINKED ISSUE #384: `\?` with argument returns error
```
(not connected)=> \? options
General
  \q                                   quit usql
  \copyright                           show usql usage and distribution terms
  \drivers                             display information about available database drivers
...
Help
  \? [commands]                        show help on backslash commands
  \? options                           show help on usql command-line options
  \? variables                         show help on special variables
...
\?: extra argument "options" ignored

(not connected)=> \? variables
...
\?: extra argument "variables" ignored
```

nineinchnick:
The `\?` command already printed help for all available commands, but it also should support subcommands for options and variables. This PR implements them and adds autocomplete.

Things to look at:
* Options are always printed to stdout, ignoring the handler's settings.
* All variable names and descriptions are kept in lists but converted to a string, it might as well be just one big string. I did **not** attempt to refactor the code so they're all defined in a single place.
* I did **not** attempt to address #379 since the README is big enough already and variables are also already referenced in a few places. I'm not convinced copying the output of `\? variables` there would help.

Resolves #384 and partially resolves #258 

BTW I noticed the release plugin gathers release notes from PRs so I'll try to avoid committing important changes directly to the default branch.

### DIFF
--- a/env/types.go
+++ b/env/types.go
@@ -16,7 +16,177 @@ import (
 	"github.com/xo/usql/text"
 )
 
-// Vars is a map of variables.
+type varName struct {
+	name string
+	desc string
+}
+
+func (v varName) String() string {
+	return fmt.Sprintf("  %s\n    %s\n", v.name, v.desc)
+}
+
+var varNames = []varName{
+	{
+		"ECHO_HIDDEN",
+		"if set, display internal queries executed by backslash commands; if set to \"noexec\", just show them without execution",
+	},
+	{
+		"ON_ERROR_STOP",
+		"stop batch execution after error",
+	},
+	{
+		"PROMPT1",
+		"specifies the standard " + text.CommandName + " prompt",
+	},
+	{
+		"QUIET",
+		"run quietly (same as -q option)",
+	},
+	{
+		"ROW_COUNT",
+		"number of rows returned or affected by last query, or 0",
+	},
+}
+
+var pvarNames = []varName{
+	{
+		"border",
+		"border style (number)",
+	},
+	{
+		"columns",
+		"target width for the wrapped format",
+	},
+	{
+		"csv_fieldsep",
+		`field separator for CSV output (default ",")`,
+	},
+	{
+		"expanded",
+		"expanded output [on, off, auto]",
+	},
+	{
+		"fieldsep",
+		`field separator for unaligned output (default "|")`,
+	},
+	{
+		"fieldsep_zero",
+		"set field separator for unaligned output to a zero byte",
+	},
+	{
+		"footer",
+		"enable or disable display of the table footer [on, off]",
+	},
+	{
+		"format",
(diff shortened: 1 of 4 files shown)


env/types.go
```diff
@@ -16,7 +16,181 @@ import (
...
+		"ECHO_HIDDEN",
+		"if set, display internal queries executed by backslash commands; if set to \"noexec\", just show them without execution",
+	},
+	{
+		"ON_ERROR_STOP",
+		"stop batch execution after error",
+	},
+	{
+		"PROMPT1",
+		"specifies the standard psql prompt",
```
kenshaw, line 39: usql

kenshaw: or, as in below, use the proper `text.*` value ...

env/types.go
```diff
@@ -16,7 +16,181 @@ import (
...
+	{
+		"SYNTAX_HL_OVERRIDE_BG",
+		"enables overriding the background color of the chroma styles",
+	},
+	{
+		"SHELL",
+		"shell used by the \\! command",
+	},
+	{
+		"TMPDIR",
```
kenshaw, line 188: I don't think this one is actually used anywhere?

metacmd/cmds.go
```diff
@@ -49,7 +50,20 @@ func init() {
...
+				if err != nil {
+					return err
+				}
+				switch name {
+				default:
+					Listing(p.Handler.IO().Stdout())
+				case "commands":
+					Listing(p.Handler.IO().Stdout())
+				case "options":
+					kingpin.Usage()
```
kenshaw, line 63: This is some bad tight coupling. It's fine for this, but let's put in a `// FIXME: decouple`.

kenshaw (changes requested):

env/types.go
kenshaw: To clarify, I mean the `TMPDIR` variable.

env/types.go
nineinchnick: Thanks, fixed.

env/types.go
nineinchnick: Removed.

metacmd/cmds.go
nineinchnick: I agree, I don't have an idea how to do it better right now. Added a comment.

Issue #389, opened 2023-01-04
go mod has error


hantmac:
I fork the code and execute `go get ./...` but got an error:

```
github.com/IBM/nzgo/v14: cannot find module providing package github.com/IBM/nzgo/v14
```

And I can not find the module `github.com/IBM/nzgo/v14` in project `github.com/IBM/nzgo` but I found `github.com/IBM/nzgo/v12` instead.

kenshaw:
I believe this is due to using an older version of Go. Closing this issue, as I have no issues with v14 of the driver.

hantmac:
> I believe this is due to using an older version of Go. Closing this issue, as I have no issues with v14 of the driver.

But I have not found the v14 driver in [this repo](github.com/IBM/nzgo).

<img width="1233" alt="image" src="https://user-images.githubusercontent.com/7600925/210729055-2133dd2c-3b2d-410b-8055-dcdb9db7d459.png">

hantmac:
@kenshaw The newest version of it is v12. Where is the 'v14' module?

nineinchnick:
Maybe the maintainers have published it by mistake and deleted it but it got cached? BTW [the docs page](https://pkg.go.dev/github.com/IBM/nzgo?tab=versions) still shows both v12 and v14.

nineinchnick:
Also these work:
```bash
% go list -m --versions github.com/IBM/nzgo/v14
github.com/IBM/nzgo/v14 v14.0.0
% go list -m --versions github.com/IBM/nzgo/v12
github.com/IBM/nzgo/v12 v12.0.1 v12.0.3 v12.0.5 v12.0.6 v12.0.7 v12.0.8
```

hantmac:
It got cached. I clone the code and can not found the module.

```
❯ go get ./...
go: github.com/IBM/nzgo/v14@v14.0.0: reading https://goproxy.io/github.com/%21i%21b%21m/nzgo/v14/@v/v14.0.0.mod: 404 Not Found
        server response: not found: github.com/IBM/nzgo/v14@v14.0.0: invalid version: unknown revision v14.0.0
❯ go list -m --versions github.com/IBM/nzgo/v14
go: github.com/IBM/nzgo/v14@v14.0.0: reading https://goproxy.io/github.com/%21i%21b%21m/nzgo/v14/@v/v14.0.0.info: 404 Not Found
        server response: not found: github.com/IBM/nzgo/v14@v14.0.0: invalid version: unknown revision v14.0.0

```

[...]

PR #390, opened 2023-01-04
fix: wrong mod version

hantmac:
Fix https://github.com/xo/usql/issues/389


kenshaw:
I have no build issues with v14. I'm guessing this is due to using an older version of Go. I'd suggest upgrading. Closing this PR, unless you can give me some kind of other concrete reason. Thanks for your work here.

PR #400, opened 2023-03-15
snowflake: enable multi-line comments

sfc-gh-pbennes:
Snowflake supports `/* */` comments, so lets enable them on the usql driver side.

Screenshot 2023-03-14 at 10 40 27 PM


sfc-gh-pbennes:
Thanks Ken!

PR #407, opened 2023-04-12
Fix FlightSQL driver

srebhan:
Not all FlightSQL backends support `SELECT version()` so we should not make any assumptions. The same holds for comment support. This PR registers an empty driver to prevent errors when connecting.

### DIFF
--- a/drivers/flightsql/flightsql.go
+++ b/drivers/flightsql/flightsql.go
@@ -4,25 +4,10 @@
 package flightsql
 
 import (
-	"context"
-
 	_ "github.com/apache/arrow/go/v12/arrow/flight/flightsql/driver" // DRIVER
 	"github.com/xo/usql/drivers"
 )
 
 func init() {
-	drivers.Register("flightsql", drivers.Driver{
-		AllowMultilineComments: true,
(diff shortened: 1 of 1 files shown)


kenshaw:
@srebhan I thought that might be the case. Instead of not doing any version string, I'd prefer if this just ignore the error and return "FlightSQL <unknown>".

kenshaw:
@srebhan for what it's worth, the only container image I could find was the `docker.io/voltrondata/flight-sql` image. If you have a better/more useful image to use, I'd gladly change my testing to using it.

srebhan:
@kenshaw currently there is no docker image available for the database I'm testing against. Will let you know once this changes.

kenshaw:
@srebhan I apologize; I forgot that I had the `drivers` package do a default version query when no `Version` hook was provided to the `drivers.Driver` definition. As such, your original submission of just `drivers.Driver{}` would have been the most correct. Sorry! It's been about 6 years since I originally wrote that, and details escaped me.

srebhan:
@kenshaw no worries! We are all humans... Except for the machines... and the animals. ;-)

Issue #235, opened 2021-04-19
Index type is always "index" in `\d` for PostgreSQL


nineinchnick:
When describing indexes using `\d` it should say what's the kind of the index, like is it a `btree` or `gist` in PostgreSQL, but it's always set to `index`.

kenshaw:
I'll check this and add to the repo if it checks out. Thanks!

PR #415, opened 2023-05-25
Fix wrong index access method in PostgreSQL \d \di slash commands

kozikowskik:
Changes related to Issue #235. These changes fix the wrong index type returned by PostgreSQL slash commands.

### DIFF
--- a/drivers/metadata/postgres/metadata.go
+++ b/drivers/metadata/postgres/metadata.go
@@ -277,11 +277,17 @@ SELECT
   c.relname as "Name",
   CASE i.indisprimary WHEN TRUE THEN 'YES' ELSE 'NO' END,
   CASE i.indisunique WHEN TRUE THEN 'YES' ELSE 'NO' END,
-  CASE c.relkind WHEN 'r' THEN 'table' WHEN 'v' THEN 'view' WHEN 'm' THEN 'materialized view' WHEN 'i' THEN 'index' WHEN 'S' THEN 'sequence' WHEN 's' THEN 'special' WHEN 'f' THEN 'foreign table' WHEN 'p' THEN 'partitioned table' WHEN 'I' THEN 'partitioned index' END as "Type"
+  COALESCE(am.amname, 
+  	CASE c.relkind 
+		WHEN 'i' THEN 'index' 
+		WHEN 'I' THEN 'partitioned index' 
+	END
+   ) as "Type"
 FROM pg_catalog.pg_class c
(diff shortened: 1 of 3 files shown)


drivers/metadata/metadata.go
```diff
@@ -466,14 +467,15 @@ func (i IndexSet) Get() *Index {
...
-	Columns   string
+	Catalog      string
+	Schema       string
+	Table        string
+	Name         string
+	IsPrimary    Bool
+	IsUnique     Bool
+	Type         string
+	Columns      string
+	AccessMethod string
```
nineinchnick, line 478: Is there a case when both index type and access method are used at the same time? If not, maybe we shouldn't add a new field here and only return access method as the type in the postgresql driver? The query that fetched this should use COALESCE

kozikowskik: I added suggested changes.

drivers/metadata/postgres/metadata.go
```diff
@@ -277,11 +277,24 @@ SELECT
   c.relname as "Name",
   CASE i.indisprimary WHEN TRUE THEN 'YES' ELSE 'NO' END,
   CASE i.indisunique WHEN TRUE THEN 'YES' ELSE 'NO' END,
-  CASE c.relkind WHEN 'r' THEN 'table' WHEN 'v' THEN 'view' WHEN 'm' THEN 'materialized view' WHEN 'i' THEN 'index' WHEN 'S' THEN 'sequence' WHEN 's' THEN 'special' WHEN 'f' THEN 'foreign table' WHEN 'p' THEN 'partitioned table' WHEN 'I' THEN 'partitioned index' END as "Type"
+  COALESCE(am.amname, 
+  	CASE c.relkind 
+		WHEN 'r' THEN 'table'
```
nineinchnick, line 282: Nit: there's a condition below to only include `relkind` of `i`, `I` and empty. Can you remove other cases here?

nineinchnick (approved):

drivers/metadata/metadata.go
```diff
@@ -466,14 +466,14 @@ func (i IndexSet) Get() *Index {
...
-	Type      string
-	Columns   string
+	Catalog      string
+	Schema       string
+	Table        string
+	Name         string
+	IsPrimary    Bool
+	IsUnique     Bool
+	Type         string
+	Columns      string
```
nineinchnick, line 476: Revert this change

PR #417, opened 2023-06-05
Add support for custom tls config in Vertica driver

vilderxyz:
Vertica driver supports [custom TLS configuration](https://github.com/vertica/vertica-sql-go#using-custom-tls-config), but the path to the custom CA needs to be passed somehow. I was considering adding it as a variable or as a connection string query. Finally, I found that passing it in queries was sufficient for me. This approach has worked for me, so I have decided to create a pull request and share my small changes.

Let me know how you feel about it. I am open to discussion and ready to make adjustments to ensure that these changes align with your requirements.

With this change, an example connection string with path to custom CA would look like this - vertica://user:password@host:port/database?tlsmode=/path/to/ca

12.06.2023:
- with those changes, secure connection with custom TLS config can be established via `vertica://<user>:<password>@<host>:<port>/<database>?tlsmode=server-strict&ca_path=/path/to/ca`, where `ca_path` is one of custom config. The solution allows to support more of [TLS options](https://pkg.go.dev/crypto/tls#Config) in the connection's string query if added to `customTlsConfig` map.

### DIFF
--- a/drivers/vertica/vertica.go
+++ b/drivers/vertica/vertica.go
@@ -5,16 +5,65 @@ package vertica
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"database/sql"
+	"fmt"
+	"io"
+	"net/url"
 	"os"
 	"regexp"
 	"strings"
 
-	_ "github.com/vertica/vertica-sql-go" // DRIVER
+	vertigo "github.com/vertica/vertica-sql-go" // DRIVER
 	"github.com/vertica/vertica-sql-go/logger"
+	"github.com/xo/dburl"
 	"github.com/xo/usql/drivers"
 )
 
 func init() {
+	// List of custom TLS configurations that may be applied via query in connection string.
+	customTlsConfig := map[string]func(string, *tls.Config) error{
+		"ca_path": func(queryValue string, c *tls.Config) error {
+			rootCertPool := x509.NewCertPool()
+
+			pem, err := os.ReadFile(queryValue)
+			if err != nil {
+				return err
+			}
+
+			if ok := rootCertPool.AppendCertsFromPEM(pem); !ok {
+				return fmt.Errorf("error: failed to append pem to cert pool")
+			}
+
+			c.RootCAs = rootCertPool
+
+			return nil
+		},
+	}
+
+	hasCustomTlsConfig := func(queries url.Values) bool {
+		for key := range customTlsConfig {
+			if queries.Has(key) {
+				return true
+			}
+		}
+
+		return false
+	}
+
+	applyCustomTlsConfig := func(queries url.Values, tlsConfig *tls.Config) error {
+		for key, configFunction := range customTlsConfig {
+			if queries.Has(key) {
+				if err := configFunction(queries.Get(key), tlsConfig); err != nil {
+					return err
+				}
+			}
+		}
+
+		return nil
+	}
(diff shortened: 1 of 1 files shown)


drivers/vertica/vertica.go
```diff
@@ -30,6 +45,41 @@ func init() {
...
+					if err != nil {
+						return nil, err
+					}
+
+					if ok := rootCertPool.AppendCertsFromPEM(pem); !ok {
+						return nil, fmt.Errorf("error: failed to append pem to cert pool")
+					}
+
+					tlsConfig := &tls.Config{RootCAs: rootCertPool, ServerName: u.Hostname()}
+					if err := vertigo.RegisterTLSConfig("custom_tls_config", tlsConfig); err != nil {
```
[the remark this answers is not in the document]
vilderxyz: You're right, I did that strictly for simplicity because I couldn't resolve the case when the user added **tlsmode** and the **path to the certificate file** separately in the query, and I wasn't sure which one should take precedence. Overall, the custom tls config acts like **server-strict**, but it also trusts self-signed certificate in provided path.

And there must be only one **tlsmode** query at the end in connection string as described
- 'none' (default) = don't use SSL/TLS for this connection
- 'server' = server must support SSL/TLS, but skip verification (INSECURE!)
- 'server-strict' = server must support SSL/TLS
- {customName} = use custom registered tls.Config (see "Using custom TLS config" section below)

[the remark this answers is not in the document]
nineinchnick: It's better to be explicit than implicit, so I'd add additional validation to check that `tlsmode=server-strict` if an additional cert is provided in a separate parameter.

drivers/vertica/vertica.go
```diff
@@ -30,6 +79,35 @@ func init() {
 			}
 			return ver, nil
 		},
+		Open: func(_ context.Context, u *dburl.URL, stdout, stderr func() io.Writer) (func(string, string) (*sql.DB, error), error) {
+			return func(_ string, params string) (*sql.DB, error) {
```
nineinchnick, line 83: Nit: since there are two values being returned here, extract this function into a named one, to make this more readable.

drivers/vertica/vertica.go
```diff
@@ -30,6 +79,35 @@ func init() {
...
+				queries := u.Query()
+
+				if queries.Get("tlsmode") == "server-strict" && hasCustomTlsConfig(queries) {
```
nineinchnick, line 86: Avoid silently ignoring parameters and return an actionable error message instead.
```suggestion
				if hasCustomTlsConfig(queries) {
				    if queries.Get("tlsmode") != "server-strict" {
				        return nil, fmt.Errorf("error: when ca_path is set, tlsmode must be set to server-strict")
				    }
```

nineinchnick (approved):
Looks good, just two nitpicks.

drivers/vertica/vertica.go
vilderxyz: I believe I cannot extract nested function form here because it uses variable **u *dburl.URL** from local scope.

vodrazka:
hi @nineinchnick, do you know when this could be merged and released?

nineinchnick (approved):

nineinchnick:
Sorry, I forgot about it. LGTM, thank you for the contribution and your patience!

PR #418, opened 2023-06-12
Alternative readline

nineinchnick:
Use an alternative readline library.

This PR includes some refactoring, all as separate commits. The last commit is the one that actually adds the new readline library.


nineinchnick:
This is a draft, since I just checked if this builds and runs. I'll be doing more manual testing now. There are some issues with the completer.

PR #423, opened 2023-08-13
Add Ingres (Vector) database support

ildus:
Add Ingres (Actian Vector) database support to usql.

There are still things to do, but basic parts like metadata for tables, indexes, columns
are working.

Linked with Ingres library through cgo so it is disabled by default.


kenshaw:
I'll review this an the `dburl` PR later today.

kenshaw:
Is there a reason you aren't using the ODBC driver with Ingress? I'm extremely hesitant to add another CGO based driver, when the ODBC connection is available for this database. Additionally, the difficulty of testing this database, auditing this code, etc. when the driver is not published by the company behind this database, really makes an extremely challenging rationale for including this.

ildus:
I just prefer using OpenAPI, it is more transparent to use and easier to set up if you have a Vector installation. ODBC is another layer on top of OpenAPI and I was confused with strange and useless error messages when I was trying to use it. That's was the reason of writing OpenAPI wrapper in Go for usql.

ildus:
I'm closing this PR. I guess if anyone will need it, they will just have to use my fork.

kenshaw:
Hi @ildus I do appreciate your contribution here. But I did make a hard decision to not add any more CGO based drivers for less well known databases. Considering there's a ODBC driver available for this, in my humble opinion `usql` already supports this. Please understand that due to the necessity of making binary releases, and limited development time for `usql`, it's a matter of practicality. There's also a small philosophical angle that everything should be pure Go where possible. I would encourage you to reach out tho the Ingres/Vector database devs and ask them for a pure Go driver. Alternately, I'd suggest trying to reverse engineer / port their JDBC driver (which I think is pure Java?) if they're unwilling to make a pure Go driver available.

PR #429, opened 2023-10-19
add alter and drop autocomplete

### LINKED ISSUE #196: Autocomplete for DDL commands
Implement autocomplete for DDL commands like `CREATE`, `ALTER`, and `DROP`.

rubiagatra:
complete #196 

Hello @nineinchnick, I believe I've created the most commonly used ALTER and DROP autocompletions. Will this close #196, or do I need to be more comprehensive about this?


drivers/completer/completer.go
```diff
@@ -353,6 +353,17 @@ func (c completer) complete(previousWords []string, text []rune) [][]rune {
 	if TailMatches(IGNORE_CASE, previousWords, "CREATE", "TABLE", "*") || TailMatches(IGNORE_CASE, previousWords, "CREATE", "TEMP|TEMPORARY", "TABLE", "*") {
 		return CompleteFromList(text, "(")
 	}
+
+	/* Complete ALTER */
+	if TailMatches(IGNORE_CASE, previousWords, "ALTER") {
```
nineinchnick, line 358: This is a good start, but would be great if this also completed table names. See https://github.com/postgres/postgres/blob/master/src/bin/psql/tab-complete.c#L1814 for reference, and also check a few lines below how we already do this for `INSERT INTO`.

Completing other objects like index names would also be great, but we can stick to tables in this PR.

PR #452, opened 2024-03-26
add column-comment

ZeyuRemtes:
column of comments plz


drivers/metadata/informationschema/metadata.go
```diff
@@ -219,6 +219,7 @@ func (s InformationSchema) Columns(f metadata.Filter) (*metadata.ColumnSet, erro
 		s.clauses[ColumnsNumericScale],
 		s.clauses[ColumnsNumericPrecRadix],
 		s.clauses[ColumnsCharOctetLength],
+		"COLUMN_COMMENT",
```
nineinchnick, line 222: This doesn't work with PostgreSQL. It needs to be optional, configurable in specific drivers, same as other clauses above:
```
% go test github.com/xo/usql/drivers/metadata/informationschema/ -cleanup=0
2024/04/13 19:16:23 Could not read pgsql columns: pq: column "column_comment" does not exist
FAIL	github.com/xo/usql/drivers/metadata/informationschema	6.947s
FAIL
```

PR #455, opened 2024-04-10
Fix Clickhouse Test

murfffi:
Fix a couple of issues in drivers/clickhouse/clickhouse_test.go

- docker/cli and docker/docker versions in go.mod had gotten out of sync and the test didn't compile
- three function aliases were expected but not present in `select name from system.functions` . This is the case either for 22.7 used by the test or the latest Clickhouse. Looking at Clickhouse bugs like https://github.com/ClickHouse/ClickHouse/issues/48260 , Clickhouse rarely includes aliases in system.functions.

`go test ./drivers/clickhouse` passes with these changes.


murfffi:
My ultimate goal is to make pr improving \copy support for Clickhouse but didn't want to include unrelated test fixes there.

murfffi:
Thanks for taking a look, @nineinchnick !

PR #457, opened 2024-04-12
Fixes in \copy implementation for Clickhouse

murfffi:
A couple of fixes in the Copy function for Clickhouse:

- Don't prepare the SELECT statement that determines columns for the destination table as the Clickhouse driver omly supports preparing INSERT statements
- De-reference the pointers populated by Scan before supplying them to arguments of Exec because Clickhouse, in some cases, doesn't accept a pointer to an arguments, instead of the argument itself. For example, this happens when source database driver doesn't support ScanType() and the value populated by Scan is interface{}.

The new automated test fails without the fixes and succeeds with them.

### DIFF
--- a/drivers/clickhouse/clickhouse.go
+++ b/drivers/clickhouse/clickhouse.go
@@ -5,7 +5,10 @@
 package clickhouse
 
 import (
+	"context"
 	"database/sql"
+	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


murfffi:
The automated test could have been either in driver_test.go or in clickhouse_test.go . The copy test in driver_test.go
expects destination DB supports RowsAffected but Clickhouse doesn't. So I made a dedicated copy test in clickhouse_test.go. I can implement the other way but the code in drivers_test.go can become convoluted.

drivers/clickhouse/clickhouse.go
```diff
@@ -35,7 +38,82 @@ func init() {
...
+		valueRefs[i] = reflect.New(columnTypes[i].ScanType())
+		values[i] = valueRefs[i].Interface()
+	}
+	var n int64
+	for rows.Next() {
+		err = rows.Scan(values...)
+		if err != nil {
+			return n, fmt.Errorf("failed to scan row: %w", err)
+		}
+		//We can't use values... in Exec() below, because, in some cases, clickhouse
```
nineinchnick, line 97: Is this something that others drivers would also benefit from?

drivers/clickhouse/clickhouse.go
```diff
@@ -35,7 +38,82 @@ func init() {
...
+				return 0, fmt.Errorf("failed to execute query to determine target table columns: %w", err)
+			}
+			columns, err := colRows.Columns()
+			_ = colRows.Close()
+			if err != nil {
+				return 0, fmt.Errorf("failed to fetch target table columns: %w", err)
+			}
+			table += "(" + strings.Join(columns, ", ") + ")"
+		}
+		// TODO if the db supports multiple rows per insert, create batches of 10 rows
```
nineinchnick, line 68: Remove TODO comments from code and create a GitHub issue instead.

drivers/clickhouse/clickhouse.go
murfffi: Yes. I think both changes apply to the general drivers.CopyWithInsert implementation. More testing will be needed though since CopyWithInsert is used in 10 drivers.

drivers/clickhouse/clickhouse.go
murfffi: Done. https://github.com/xo/usql/issues/458 .

drivers/clickhouse/clickhouse.go
murfffi: How about doing this in three PRs

1. This PR fixes Clickhouse only at first
2. I do https://github.com/xo/usql/issues/458 for Clickhouse
3. Replace drivers.CopyWithInsert with this implementation + optional batching + the placeholder code currently in drivers.CopyWithInsert . Extend TestCopy drivers_test.go to include a source DB that doesn't support ScanType (e.g. csvq), and a target DB that only support prepared INSERT statements (e.g. Clickhouse, Databend).

Trying to go directly to step 3 will delay the fix for \copy for Clickhouse a lot .

Issue #478, opened 2024-08-09
table name tab-completion not working between 0.18.x -> 0.19.x


tomans-spirent:
Description of issue:

When I upgrade from the 0.18.x release to the 0.19.x release the tab-completion of the table names in places like `select * from` no longer works with no errors shown. 

Background context:

I am running usql_static on ubuntu 20.04 amd64, connecting to a postgres 12 instance. 

Something between those two versions might have broken this functionality.

remisalmon:
Same issue here, https://github.com/xo/usql/releases/tag/v0.18.1 is the last version where the tab-completion works when connecting to snowflake.

PrayagS:
+1. Autocompletion for `\d*` commands is not working with postgres in version `0.19.12`.

jorgheymans:
Same for Oracle. Pretty annoying as with a TUI autocomplete is a rather substantial part of the productivity :-(

remisalmon:
This got fixed in https://github.com/xo/usql/commit/78bd36718e38ff1199f24bf9a12b5a41ce048892 that is in the v0.20.1 release.

PR #481, opened 2024-09-25
Fix drivers completions

remisalmon:
Fixes https://github.com/xo/usql/issues/478

The drivers NewCompleter was removed from the handler between v0.18.1 and v0.19.0.
Tested locally, I had to cleanup the dependencies for go to build.


handler/handler.go
```diff
@@ -783,6 +783,7 @@ func (h *Handler) Open(ctx context.Context, params ...string) error {
 	// force error/check connection
 	if err == nil {
 		if err = drivers.Ping(ctx, h.u, h.db); err == nil {
+			h.l.Completer(drivers.NewCompleter(ctx, h.u, h.db, readerOpts(), completer.WithConnStrings(h.connStrings())))
```
nineinchnick, line 786: Add a check for `h.l.Interactive()` before calling this.

nineinchnick (commented):
Thanks a lot for working on it! I found that the completer calls were moved from `Open` to `New` in https://github.com/xo/usql/commit/05236d14c834544688af136a0bd9a3b4909798f9. We should figure out why, to avoid calling it twice. @kenshaw do you have more context why? Should we revert this one change?

handler/handler.go
remisalmon: Done in https://github.com/xo/usql/pull/481/commits/c0940a3a766548f81f4915edf9d6284a867d8e1a.

remisalmon:
> Thanks a lot for working on it! I found that the completer calls were moved from `Open` to `New` in [05236d1](https://github.com/xo/usql/commit/05236d14c834544688af136a0bd9a3b4909798f9). We should figure out why, to avoid calling it twice. @kenshaw do you have more context why? Should we revert this one change?

I was thinking to also move the drivers.NewCompleter() call to `New` but `New` is missing the context that this function needs.
I moved all completer calls back to `Open` in https://github.com/xo/usql/pull/481/commits/2ad11cf41940e3fb14f5246e0c63796e5d45d286 to simplify a little but I can revert if you prefer?

handler/handler.go
```diff
@@ -803,6 +799,11 @@ func (h *Handler) Open(ctx context.Context, params ...string) error {
 	// force error/check connection
 	if err == nil {
 		if err = drivers.Ping(ctx, h.u, h.db); err == nil {
+			if h.l.Interactive() {
+				h.l.SetOutput(h.outputHighlighter)
+				h.l.Completer(completer.NewDefaultCompleter(completer.WithConnStrings(h.connStrings())))
```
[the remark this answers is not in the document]
remisalmon: Reverted that in https://github.com/xo/usql/pull/481/changes/14d1f3020fd730c3e1cad75d3d61194c7fd2bb97.

kenshaw:
@remisalmon Hi, I apologize for not looking at this earlier. Could you squash the commits into a single commit and I'll rebase onto master and backport it to the `release-20` branch? Thanks.

Issue #458, opened 2024-04-12
Batch insert in \copy to Clickhouse DB destination


murfffi:
Clickhouse supports [multi-row INSERT statements](https://clickhouse.com/docs/en/sql-reference/statements/insert-into) so 
it is possible to speed up \copy by using batches. This issue is discussed in the PR for Clickhouse \copy implementation -
https://github.com/xo/usql/pull/457 .

murfffi:
I volunteer to do this after https://github.com/xo/usql/pull/457 .

PR #501, opened 2024-12-02
bulk load for MySQL

murfffi:
The PR implements bulk loading for MySQL using the "LOAD DATA from io.Reader" feature of github.com/go-sql-driver/mysql - https://github.com/go-sql-driver/mysql?tab=readme-ov-file#load-data-local-infile-support .
As expected, bulk loading this way is significantly faster. 1 mln. rows in the "staff" table from the test schema are inserted for 15 sec vs. 120 sec using INSERT: **8x improvement**. Note that LOAD DATA INFILE LOCAL is disabled by default on MySQL 8+ servers and must be enabled using SET GLOBAL local_infile = ON beforehand. MySQL doesn't seem to have any remote bulk loading options that are enabled by default. The implementation detects if bulk copy is enabled on the server and falls back to regular copy if needed.

The PR also extends TestCopy in drivers_test.go with comparison of copied data to ensure MySQL bulk loading is safe across data types.

Testing Done: tests in drivers_test.go

### DIFF
--- a/drivers/drivers_test.go
+++ b/drivers/drivers_test.go
@@ -7,11 +7,13 @@ import (
 	"bytes"
 	"context"
 	"database/sql"
+	"errors"
 	"flag"
 	"fmt"
 	"log"
 	"net/url"
 	"os"
+	"reflect"
 	"regexp"
 	"strings"
 	"testing"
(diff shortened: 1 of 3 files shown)


murfffi:
This is ready for review.

Background: I recently resumed working on #458 and while looking for a database to experiment with generic bulk copy, I saw that MySQL supports bulk load similar to Postgres. I decided to implement that first.

murfffi:
The PR checks failure seems unrelated. The checks on master fail the same way. I got the checks to pass in https://github.com/xo/usql/pull/503 but that PR probably can't be merged as is.

murfffi:
Just a reminder that I am available to merge this if needed. Let me know if I should rebase to release-21.

I also edited the description to clarify that the implementation will fall back to regular copy if bulk copy is not enabled on the server.

PR #503, opened 2024-12-07
Fix substitution of quoted variables 

### LINKED ISSUE #505: regression: quoted variables replacement doesn't work on master
The following works on the latest release v0.19.14 but fails on master. This is also why the PR checks fail:

image

Fixed in PR #503 .

murfffi:
testcli.go failed before the fix and passes after it.
Also includes a fix to testcli.go itself.

Closes #505 .

### DIFF
--- a/testcli.go
+++ b/testcli.go
@@ -12,6 +12,7 @@ import (
 	"log"
 	"os"
 	"regexp"
+	"strings"
 	"time"
 
 	gexpect "github.com/google/goexpect"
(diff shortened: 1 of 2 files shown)


murfffi:
Please advise if a new unit test is needed here. stmt_test.go contains tests for statement parsing (TestNextResetState) and substitution (TestVarSubstitute) but I didn't see a test for both at the same time. (I might have missed it - there is a lot of tests.) Maybe this is where testcli.go comes in though it took me some time to get from the testcli.go failure to the root cause.

testcli.go
```diff
@@ -127,9 +128,18 @@ func (test Test) do(ctx context.Context, binpath string, timeout time.Duration)
...
 	}
 	for _, line := range bytes.Split(buf, []byte{'\n'}) {
+		// give time to usql to react to sent commands so statements are close to their results in output
+		// this helps link failures to the statement that caused them
+		time.Sleep(100 * time.Millisecond)
 		if err := exp.Send(string(line) + "\n"); err != nil {
 			return err
 		}
+		command := strings.TrimSpace(string(line))
+		if command == "\\q" || command == "\\quit" {
```
ccoVeille, line 138: String literal can help readability

```suggestion
		if command ==`\q` || command == `\quit` {
```

PR #510, opened 2025-01-25
Add default filters for sqlite3 metadata tables reader

### LINKED ISSUE #504: `\dt` unexpectedly lists SQLite system tables
In sqlite, `\dt` is intended to list only user-created tables, but it also includes system tables.

#### 1. prepare a test.db

Create a `test.sql` file
```sql 
create table test (test_id int, name string);
insert into test (test_id, name) values (1, 'hello');
```

Execute the sql file
```
usql sqlite3://test.db -f test.sql
```

#### 2. run `\dt` on the test.db
```
$ usql sqlite3://test.db -c '\dt'
             List of relations
 Schema |        Name        |     Type     
--------+--------------------+--------------
        | fts3               | SYSTEM TABLE 
        | fts3tokenize       | SYSTEM TABLE 
        | fts4               | SYSTEM TABLE 
        | fts4aux            | SYSTEM TABLE 
        | fts5               | SYSTEM TABLE 
        | fts5vocab          | SYSTEM TABLE 
        | json_each          | SYSTEM TABLE 
        | json_tree          | SYSTEM TABLE 
        | pragma_module_list | SYSTEM TABLE 
        | rtree              | SYSTEM TABLE 
        | rtree_i32          | SYSTEM TABLE 
        | test               | TABLE 
(12 rows)
```

### Additional Context

```
$ usql --version
usql 0.19.14
```

kozikowskik:
The current response from the metacmd `\dt` and `\dv' in `sqlite3` contains system tables, which is not expected. These changes add table types allowed in the result of the above metacmds.

Fixes #504

### DIFF
--- a/drivers/sqlite3/sqshared/reader.go
+++ b/drivers/sqlite3/sqshared/reader.go
@@ -3,6 +3,7 @@ package sqshared
 import (
 	"database/sql"
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/xo/usql/drivers"
@@ -125,8 +126,18 @@ FROM (
 		conds = append(conds, "table_name LIKE ?")
 	}
 	if len(f.Types) != 0 {
+		tableTypes := []string{
+			"BASE TABLE",
+			"TABLE",
+			"VIEW",
+			"GLOBAL TEMPORARY",
+		}
 		pholders := []string{}
 		for _, t := range f.Types {
+			if !slices.Contains(tableTypes, t) {
+				continue
+			}
+
 			vals = append(vals, t)
 			pholders = append(pholders, "?")
 		}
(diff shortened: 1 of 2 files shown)


drivers/sqlite3/sqshared/reader.go
```diff
@@ -125,8 +126,18 @@ FROM (
...
 	if len(f.Types) != 0 {
+		tableTypes := []string{
+			"BASE TABLE",
+			"TABLE",
+			"VIEW",
+			"GLOBAL TEMPORARY",
+		}
 		pholders := []string{}
 		for _, t := range f.Types {
+			if !slices.Contains(tableTypes, t) {
```
nineinchnick, line 137: This prevents from getting system tables at all. We should check `f.WithSystem` here.

nineinchnick:
We talked offline and decided to open another PR.

Issue #511, opened 2025-02-07
Help to connect to CosmosDB


rodfersou:
Hello folks,

I'm trying to connect to CosmosDB without luck:

```bash
~/projects/folder   
$ usql cm://$CONNECTIONSTRING 
error: missing user

~/projects/folder 
$ usql cm://$USER@$CONNECTIONSTRING
error: cosmos: sql: unknown driver "cosmos" (forgotten import?)
```

Knowing that my connection string follow  [gocosmos](https://github.com/btnguyen2k/gocosmos) DSN:

```
"AccountEndpoint=https://localhost:8081/;AccountKey=<cosmosdb-account-key>"
```

Thanks in advance 🙏

rodfersou:
@kenshaw can you help me here please? 🙏

murfffi:
There is a small bug in https://github.com/xo/usql/blob/master/drivers/cosmos/cosmos.go - driver name is `gocosmos` , not `cosmos` . Ref: https://github.com/btnguyen2k/gocosmos?tab=readme-ov-file#example-usage .

rodfersou:
@murfffi user should be optional, but right now is required

murfffi:
Hey @rodfersou , `usql` uses the [dburl](https://github.com/xo/dburl/blob/master/dsn.go#L181) library to convert URLs passed on the command-line to gocosmos DSN. The URL user is used as AccountKey. The https://github.com/btnguyen2k/gocosmos readme states that AccountKey is required so URL user is required. [(code)](https://github.com/xo/dburl/blob/master/dsn.go#L181)

If you want to use the DSN in the format described at https://github.com/btnguyen2k/gocosmos , then use the `\connect` command instead: 

start `usql` without an URL and then run:

```shell
\c gocosmos DSN
```

In that case, you need to use the driver name - `gocosmos` , not any alias like `cm` or `cosmos` .

rodfersou:
I'm passing AccountKey, but usql is requiring `user@` before the rest of the connection string

rodfersou:
error: missing user

murfffi:
> I'm passing AccountKey, but usql is requiring `user@` before the rest of the connection string

Could you add the specific URL that you want to work? Replace sensitive parts with xxx . Maybe something like:

`cm://endpoint?AccountKey=xxxx&DefaultDb=xxx` ?

rodfersou:
```
[...]

PR #514, opened 2025-03-01
fix: strip trailing semicolon in cosmosdb and unify implementation 

murfffi:
Fixes, in part, #511 .

cosmosdb, like several other databases, needs trailing semicolons to be
removed from the statement before it is passed to the driver.

This fix applies that to cosmosdb and makes the other databases with identical
processing use the same implementation - presto, trino, athena, SAP ASE.

Drivers that strip trailing semicolon as part of other processing are
not included in the refactoring to keep scope small.

Testing: with https://github.com/xo/dburl/pull/44 applied , testing cosmosdb ...

image

Also tested with trino.

### DIFF
--- a/drivers/athena/athena.go
+++ b/drivers/athena/athena.go
@@ -5,22 +5,15 @@ package athena
 
 import (
 	"context"
-	"regexp"
 
 	_ "github.com/uber/athenadriver/go" // DRIVER: awsathena
-	"github.com/xo/dburl"
 	"github.com/xo/usql/drivers"
 )
 
 func init() {
-	endRE := regexp.MustCompile(`;?\s*$`)
 	drivers.Register("awsathena", drivers.Driver{
 		AllowMultilineComments: true,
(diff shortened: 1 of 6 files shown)


kenshaw:
Please drop the `gocosmos` alias change, and rebase this as a single commit. I'll rebase this on `master` and back-port this to the `release-19` branch and cut a release. I'll also update the `dburl` dependency on the `release-19` branch at that time, so the `gocosmos` alias works properly. Thanks.

murfffi:
Thanks for taking a look! Squashed the commits. I left the gocosmos alias in because the two argument form of \connect - with native DSN - needs that alias to work:

image

Without native DSN \connect, users need to url-encode the cosmos API key so dburl can parse the connection string properly. I could add the alias as a separate PR but that will require even more work merging to release-19.

murfffi:
Tested trino also to confirm that the mild refactoring didn't break the other affected DBs.

PR #518, opened 2025-04-06
feat: Bring back Impala support

### LINKED ISSUE #517: feat: Impala support
It will be great if Impala support is brought back using the https://github.com/sclgo/impala-go driver that I maintain. I have a soft spot for Impala after working with it more than 5 years with it and I still think that there is a niche for it - faster than Presto, more deployment flexibility than DuckDB, more mature than Doris. 

There are no Impala CLIs besides the official one, which it is terrible - a lot of pain to install and very lacking in features.
usql support for Impala will really help everybody who (still) uses it.

I have a draft PR which I can submit if the feature makes sense. Here it is in action:

murfffi:
The PR brings back Impala support as discussed in #517 

- Changes the Impala driver from the abandoned bippio/go-impala to the maintained sclgo/impala-go.
- Moves Impala from the "bad" to the "most" driver set
- Add MetaReader for Impala
- Implements copy for Impala, extending the generic CopyWithInsert with an option to work without transactions - for DBs that don't support them
- Implements IsPasswordErr

The PR targets release-19, not master, so that the CI confirms no regressions caused by the change in drivers.CopyWithInsert. (CI on master is broken). The commit shouldn't create issues when it is cherry-picked to master afterwards.

I tested for regressions in CopyWithInsert with

```
go test -tags most -v -run=TestCopy ./drivers -args -cleanup=false -dbs=pgsql,mysql,pgx,csvq
```

Closes #517


kenshaw:
@murfffi you are the best. I'll review when I'm out of bed and add first thing.

murfffi:
Thank you!

PR #519, opened 2025-04-11
Fix all actual typos and use .codespellrc to skip others

ottok:
Fix all spelling errors found by codespell[1] version 2.4.1.

No changes to `"ALTER HEIRARCHY"` in `drivers/qtype.go`.

[1] https://github.com/codespell-project/codespell/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1468,7 +1468,7 @@ pg:booktest@=>
 ```
 
 While the `.usqlpass` functionality will not be removed, it is recommended to
-[define named connections][connection-vars] preferrably via [the `config.yaml`
+[define named connections][connection-vars] preferably via [the `config.yaml`
 file][config].
 
 <hr/>
--- a/drivers/metadata/informationschema/metadata.go
+++ b/drivers/metadata/informationschema/metadata.go
@@ -94,7 +94,7 @@ func New(opts ...metadata.ReaderOption) func(drivers.DB, ...metadata.ReaderOptio
 		systemSchemas:     []string{"information_schema"},
 		dataTypeFormatter: func(col metadata.Column) string { return col.DataType },
 	}
-	// aply InformationSchema specific options
+	// apply InformationSchema specific options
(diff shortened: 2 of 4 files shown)


kenshaw:
I appreciate the work done to find typos, but as there's only one in the `README.md`, and the others are trivial, I don't really know what the point of merging this is. This is also a Go project -- as such, I see no reason to add a configuration file to the root that is longer than all the other changes combined. I understand that this is the configuration for the tool you used, but I would never run that tool again on the codebase, nor do I believe anyone else would.

ottok:
Would you like me to remove the `.codespellrc` and just submit the typofixes?

Lintian in Debian was complaining about typofixes, so I had a need to fix them.

ccoVeille:
I agree with @kenshaw having a configuration file for a tool that won't be used later, or only by a few people is somehow a problem.

But, while I state with the same things, my conclusion and suggestion would be different.

I would like to suggest to add the GHA codespell-action to the CI

https://github.com/codespell-project/actions-codespell

I often set it up when I also jump into a project to fix typos.

So here, I would agree with the idea to either have:
- a PR that only fixes the typos and don't commit the codespell configuration file
- a PR that fixes the typos, includes the codespell configuration file, and sets up the GHA codespell-action
- one PR with the typos fixes and another PR to set up the GHA and so include the configuration file

ottok:
I already updated this PR to have the typofixes only. I don't use GitHub code actions myself, so I won't invest time in building one. Feel free to use the .codespellrc from a previous version in this PR if you later need it.

ccoVeille:
This is a good compromise.

@ottok please let me know if you have interest for adding such GitHub action

ottok:
I don't use GitHub code actions myself, so I won't invest time in building one.

ccoVeille:
I should have been clearer. I was suggesting I could help if @kenshaw wants me to look at it.

PR #524, opened 2025-05-02
fix(oracle): use view available for all users for driver.Version

murfffi:
Use v$version instead of v$instance because it is available to all users.
Like v$instance, v$version is available in all non-ancient oracle versions.
Both are available in the earliest documentation I could find: 10g, released 2003.

Testing Done: Oracle Express 21c (as in the reported issue) and 11g

For 11g:

docker run -d -p 1521:1521 -e ORACLE_PASSWORD=foobar gvenzl/oracle-xe:11-slim ./usql oracle://sys:foobar@localhost/xe
# version is shown correctly
create user test identified by test;
GRANT CREATE SESSION to test;
\connect oracle://test:test@localhost/xe
# version is still shown correctly, despite minimal privileges

21c in the issue was also tested that way.


drivers/oracle/orshared/orshared.go
```diff
@@ -37,10 +37,12 @@ func Register(name string, err func(error) (string, string), isPasswordErr func(
...
 		Version: func(ctx context.Context, db drivers.DB) (string, error) {
 			var ver string
-			if err := db.QueryRowContext(ctx, `SELECT version FROM v$instance`).Scan(&ver); err != nil {
+			// this query must work for the "public" role - the minimum privileges for a user
+			// v$version view is available at least since Oracle 10g (2003)
+			if err := db.QueryRowContext(ctx, `SELECT banner FROM v$version`).Scan(&ver); err != nil {
 				return "", err
 			}
-			return "Oracle Database " + ver, nil
+			return ver, nil // typical output: Oracle Database 21c Express Edition Release 21.0.0.0.0 - Production
```
kenshaw, line 45: Please remove the comments, they're 3x longer than the actual code.

kenshaw:
@murfffi can you please also modify `contrib/oracle*/usql-config` (there are two directories) with this revised query?

Issue #528, opened 2025-05-17
Consider using readline library from github.com/chzyer/readline


ottok:
Currently usql uses the Readline library from https://github.com/gohxs/readline, which is a fork of https://github.com/chzyer/readline.

However, gohxs/readline hasn't had any commits for 8 years. Seems to me the original library chzyer/readline would be a better choice these days, as it continues to be actively maintained? If it is missing something, perhaps take the code from gohxs/readline you rely on and send as Pull Request to chzyer/readline?

ottok:
In https://github.com/xo/usql/pull/535#issuecomment-3120838982 Ken replied:

> I just tested this, it doesn't work with syntax highlighting, which basically makes this a no-go, as I personally made this project for 2 reasons: 1) to have a tool to use for all SQL databases with the [`xo/dbtpl`](https://github.com/xo/dbtpl) project, and 2) syntax highlighting, as `psql` (and every other native SQL CLI I've used), doesn't support syntax highlighting.

PR #535, opened 2025-07-26
Replace unmaintained gohxs/readline with latest chzyer/readline

ottok:
The Go library `gohxs/readline` has no commits for past 8 years. It was originally forked from `chzyer/readline`, which continues to be maintained, so switch back to it to ensure continued maintenance and updates.

This change necessitated adapting the output filtering mechanism due to API incompatibilities that manifested as errors:

    rline/rline.go:128:16: l.Inst.Config.Output undefined (type *readline.Config has no field or method Output)

    rline/rline.go:130:7: cfg.FuncFilterOutput undefined (type *readline.Config has no field or method FuncFilterOutput)

Specifically, the `readline.Config` no longer provides direct output fields or filter functions. To address this, implement a `filterWriter` to wrap the standard output stream, ensuring existing output filtering functionality is preserved.

Closes: #528


ottok:
CI failed without error messages:
```
  go run testcli.go &> output.log
  ls -alh output.log
  shell: /usr/bin/bash -e {0}
Error: Process completed with exit code 1.
```

kenshaw:
I just tested this, it doesn't work with syntax highlighting, which basically makes this a no-go, as I personally made this project for 2 reasons: 1) to have a tool to use for all SQL databases with the [`xo/dbtpl`](https://github.com/xo/dbtpl) project, and 2) syntax highlighting, as `psql` (and every other native SQL CLI I've used), doesn't support syntax highlighting.

rline/rline.go
```diff
@@ -123,9 +125,45 @@ func (l *Rline) Password(prompt string) (string, error) {
...
+	cfg := l.Inst.Config.Clone() // Get a mutable copy of the config
+
+	var currentWriter io.WriteCloser
+	if f != nil {
+		// Wrap the original stdout with the filter
+		currentWriter = &filterWriter{writer: l.originalStdout, filter: f}
+	} else {
+		// If filter is nil, revert to the original stdout
+		currentWriter = l.originalStdout
+	}
```
ccoVeille, line 161: I tend to write this this way


```suggestion
	var currentWriter io.WriteCloser

	// use original stdout by default
	currentWriter = l.originalStdout
	if f != nil {
		// Wrap the original stdout with the filter
		currentWriter = &filterWriter{writer: l.originalStdout, filter: f}
	}
```


What to you think?

ottok:
Thanks @ccoVeille for the suggestion, but based on Ken's comment I think we need to have confirmation on the syntax highlightning issue https://github.com/chzyer/readline/issues/253 before it makes sense to polish the syntax on this change, as otherwise it won't be accepted anyway.

Issue #495, opened 2024-11-15
Invalid database scheme with `\copy`


bobvanderlinden:
When doing the following:

```
=> \cset MSSQL mssql://USER:PASS@HOST:PORT/DATABASE
=> \copy MSSQL csvq://. 'YYY'
error: invalid database scheme
```

Note that `\c MSSQL` works fine, as does `\c csvq://.`.
What could be the problem?

PR #542, opened 2025-11-25
Add conntection variables support to \copy

ejiektpobehuk:
This PR addresses #495 and similar issues I've experienced.

`\copy` had no support for connection variables set with `\cset` or a configuration file.
It went straight to parsing the source and destination with `dburl.Parse()`.

I've copied the `Handler.Open()` behaviour without any DRY.
Now it can handle connection variables.

### DIFF
--- a/metacmd/cmds.go
+++ b/metacmd/cmds.go
@@ -573,17 +573,45 @@ func Copy(p *Params) error {
 	if err != nil {
 		return err
 	}
-	src, err := dburl.Parse(srcstr)
-	if err != nil {
-		return err
+	// resolve source connection variable if needed
+	srcParams := []string{srcstr}
+	if v, ok := env.Vars().GetConn(srcstr); ok {
+		srcParams = v
+	}
+	// parse or construct source URL
+	var src *dburl.URL
+	if len(srcParams) < 2 {
+		src, err = dburl.Parse(srcParams[0])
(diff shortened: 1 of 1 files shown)


ejiektpobehuk:
I'd like to add a test to the test suite. My current version is file system dependant and I'm not sure if it's a good fit. On the other hand this test uses csvq and sqlite3 which are built-in and don't require external databases.

```
-- Create source data
\! mkdir -p /tmp/usql_test
\! echo "id,name,value" > /tmp/usql_test/source.csv
\! echo "1,Alice,100" >> /tmp/usql_test/source.csv
\! echo "2,Bob,200" >> /tmp/usql_test/source.csv
\! echo "3,Charlie,300" >> /tmp/usql_test/source.csv

-- Set up connection variables using \cset
\cset SOURCE csvq:///tmp/usql_test
\cset DEST sq:/tmp/usql_test/dest.db

-- Create destination table
\c DEST
CREATE TABLE test_table (
  id INTEGER PRIMARY KEY,
  name TEXT,
  value INTEGER
);

-- Test target
\copy SOURCE DEST 'SELECT * FROM `source.csv`' 'test_table'

-- Verify the data was copied
\c DEST
SELECT * FROM test_table;

-- Cleanup
\! rm -rf /tmp/usql_test

\quit
```

PR #555, opened 2026-03-21
fix: wrap routine_definition in COALESCE to handle NULL

### LINKED ISSUE #554: Tab completion error: Scan error on routine_definition NULL column
Pressing Tab for completion on PostgreSQL returns an error before showing results:

Error getting selectables sql: Scan error on column index 6, name "routine_definition": converting NULL to string is unsupported

Completion still works but the error is noisy.

**Environment:**
- usql latest
- PostgreSQL

GourangaDasSamrat:
Fixes #554 

`routine_definition` in `information_schema.routines` can be NULL for some function types (e.g. external/C functions). Scanning NULL into a plain `string` causes a scan error during tab completion.

Wrapping in `COALESCE` is consistent with how other nullable columns in the same query are already handled (`routine_type`, `data_type`).


kenshaw:
Added to `main`. Thanks!

GourangaDasSamrat:
@kenshaw Hey, thanks a lot for applying the fix — really appreciate it!

I noticed the changes were added manually instead of merging the PR. Would it be possible to preserve authorship in such cases (e.g., via merge or co-author)? It would really help with contribution tracking on my end 🙂

Thanks again!

kenshaw:
Apologies, I used `cherry-pick` here as I needed to add it to the `release-21` branch as well as `main`. I should have used rebase onto `main`, and then cherry-picked it to the release branch. If you're worried about attribution, the Git author and committer is still preserved, it will be visible on your GitHub commit history as far as I'm aware.

PR #572, opened 2026-05-05
snowflake: allow $$ delimited strings

jmccarv:
Enable support for $$ delimited string literals in the snowflake driver.

This enables queries that would otherwise fail, like:

`select $$Hello 'world';$$;`

### DIFF
--- a/drivers/snowflake/snowflake.go
+++ b/drivers/snowflake/snowflake.go
@@ -29,6 +29,7 @@ func init() {
 	)
 	drivers.Register("snowflake", drivers.Driver{
 		AllowMultilineComments: true,
+		AllowDollar:            true,
 		Err: func(err error) (string, string) {
 			if e, ok := err.(*gosnowflake.SnowflakeError); ok {
 				return strconv.Itoa(e.Number), e.Message


kenshaw:
@jmccarv Thanks for the contribution, I'll add to the next release. I was not aware that Snowflake support `$$` strings. Though from a quick reading it doesn't seem to support the `$tag$ ... $tag$` style [that PostgreSQL does](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING).

jmccarv:
@kenshaw Thanks for the quick response!

You're right about it not supporting the `$tag$` style. I didn't see a way to enable `$$` strings without also enabling `$tag$` in the code, but I suppose it shouldn't hurt to enable it anyway.

For giggles I tested `$tag` in snowflake with my patched version of usql:
```
=> select $xx$hi;$xx$;
error: snowflake: 1003: SQL compilation error:
syntax error line 1 at position 10 unexpected '$hi'.
```

I think that's nicer than the current error:
```
=> select $xx$hi;$xx$;
error: snowflake: 1003: SQL compilation error:
syntax error line 1 at position 10 unexpected '$hi'.
error: snowflake: 1003: SQL compilation error:
syntax error line 1 at position 0 unexpected '$xx'.
```

kenshaw:
Yeah, I agree -- the goal of `usql` is not to work perfectly with every database, just to, ya know, provide something like an easier time for people who are used to `psql` to work with every other database. It wasn't a barrier for me merging the PR, just was pointing it out. Quick -- someone go file a ticket on the Snowflake projects to get proper PostgreSQL tag support! (heh)

